PR #2, opened 2024-07-08
feat: [UEPR-15] Updated script to fix build issues and refactored logic for merging branches

MiroslavDionisiev:
### Jira Ticket
[UEPR-15](https://scratchfoundation.atlassian.net/browse/UEPR-15)

### Proposed Changes
Changed the logic for creating monorepo branches and merging code. Added logic that fixes wrong paths after the merge in the monorepo and ensures everything builds properly.

[UEPR-15]: https://scratchfoundation.atlassian.net/browse/UEPR-15?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/build-monorepo.sh
+++ b/scripts/build-monorepo.sh
@@ -469,8 +469,8 @@ process_workspace_webpack_config () {
 
 use_node_version_from_nvmrc () {
     source ~/.nvm/nvm.sh
-    nvm install < .nvmrc
-    nvm use < .nvmrc
+    nvm install
+    nvm use
 }
 
 ### Do the things! ###
--- a/scripts/build-monorepo.sh
+++ b/scripts/build-monorepo.sh
@@ -19,7 +19,6 @@ DEST_BRANCHES="
     main \
     scratch-android \
     scratch-desktop \
-    test-integration \
 "
 #ALL_REPOS="scratch-audio \
 #     scratch-desktop \
--- a/scripts/build-monorepo.sh
+++ b/scripts/build-monorepo.sh
@@ -8,26 +8,24 @@ GITHUB_ORG="scratchfoundation"
 # This is the list of repositories to merge into the monorepo
 # Current thinking: this should be all Scratch Editor repos excluding forks
 ALL_REPOS="
-    scratch-audio \
-    scratch-desktop \
     scratch-gui \
-    scratch-l10n \
-    scratch-paint \
-    scratch-parser \
     scratch-render \
-    scratch-sb1-converter \
-    scratch-semantic-release-config \
-    scratch-storage \
     scratch-svg-renderer \
-    scratch-translate-extension-languages \
     scratch-vm \
-    eslint-config-scratch \
+"
+
+DEST_BRANCHES="
+    develop \
+    main \
+    scratch-android \
+    scratch-desktop \
+    test-integration \
 "
 #ALL_REPOS="scratch-audio eslint-config-scratch"
 
 # This is the directory where you have a copy of all the repositories you want to merge.
 # This script will run `git fetch` on these repos, but otherwise will not modify them.
-BUILD_CACHE="./monorepo.cache"
+BUILD_CACHE="./.."
 
 # The monorepo will be built here. Delete it to start over.
 BUILD_OUT="./monorepo.out"
@@ -56,7 +54,7 @@ GIT_PACK_WINDOW_MEMORY="512m"
 
 set -e
 
-if ! git filter-repo --help &> /dev/null; then
+if ! git filter-repo -h &> /dev/null; then
     echo "Please install git-filter-repo. One of these commands might work:"
     echo "- brew install git-filter-repo"
     echo "- sudo apt install git-filter-repo"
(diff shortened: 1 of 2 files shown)


scripts/build-monorepo.sh
```diff
@@ -8,26 +8,24 @@ GITHUB_ORG="scratchfoundation"
...
-    scratch-l10n \
-    scratch-paint \
-    scratch-parser \
     scratch-render \
-    scratch-sb1-converter \
-    scratch-semantic-release-config \
-    scratch-storage \
     scratch-svg-renderer \
-    scratch-translate-extension-languages \
     scratch-vm \
```
sashodada, line 14: Even though they are tracked externally, should we leave the removed repos in a comment somewhere around here for future reference? There is a commented out definition of `ALL_REPOS` below we could probably reuse. Maybe also changing the comment above the `ALL_REPOS` definition to fit the current state of the script would be helpful as well

scripts/build-monorepo.sh
```diff
@@ -238,6 +214,24 @@ optimize_git_repo () {
     #du -sh "$BUILD_OUT"
 }
 
+default_branch () {
+    BRANCH=""
+    
+    if [ -z "$(git -C "${BUILD_TMP}/${REPO_NAME}" branch --list "$BRANCH")" ]; then
```
sashodada, line 288: Can this conditional be skipped since the first check is against an empty branch name?

scripts/build-monorepo.sh
```diff
@@ -295,19 +289,19 @@ fixup_branch () {
...
                 DEVDEPS="$DEVDEPS $DEP@*"
             fi
             if jq -e .optionalDependencies.\"$DEP\" "${BUILD_OUT}/workspaces/${REPO}/package.json" > /dev/null; then
-                REMOVEDEPS="$REMOVEDEPS $DEP"
+                jq  "del(.optionalDependencies.\"$DEP\")" "${BUILD_OUT}/workspaces/${REPO}/package.json"  | sponge "${BUILD_OUT}/workspaces/${REPO}/package.json"
                 OPTDEPS="$OPTDEPS $DEP@*"
             fi
             if jq -e .peerDependencies.\"$DEP\" "${BUILD_OUT}/workspaces/${REPO}/package.json" > /dev/null; then
-                REMOVEDEPS="$REMOVEDEPS $DEP"
+                jq  "del(.peerDependencies.\"$DEP\")" "${BUILD_OUT}/workspaces/${REPO}/package.json"  | sponge "${BUILD_OUT}/workspaces/${REPO}/package.json"
```
sashodada, line 370: Are the duplications in these statements worth extracting in a variable?

sashodada (approved):

scripts/build-monorepo.sh
```diff
@@ -206,6 +178,45 @@ add_repo_to_monorepo () {
...
 }
 
+add_gh_pages () {
+    REPO_NAME="$1"
+    ORG_AND_REPO_NAME="${GITHUB_ORG}/${REPO_NAME}"
+    echo "Working on $ORG_AND_REPO_NAME"
+
+    GH_PAGES_BRANCH="gh-pages"
+
+    clone_repository $REPO_NAME
```
cwillisf, line 189: Is there a way to skip this clone step, since we will already have cloned the repo in `add_repo_to_monorepo`? Or is there a reason it needs to be cloned again?

I'm wondering if `add_gh_pages` could instead be a special branch handler in the `DEST_BRANCH` loop in `add_repo_to_monorepo`. Maybe something like:
```sh
gh-pages)
  add_gh_pages $REPO_NAME
  continue
  ;;
```

cwillisf: ...or possibly even just an "if the branch is `gh-pages` then create it a little bit differently" clause in the `create the destination branch if it doesn't exist` section of `add_repo_to_monorepo`?

scripts/build-monorepo.sh
```diff
@@ -206,6 +178,45 @@ add_repo_to_monorepo () {
...
+
+    if [ -z "$(git -C "$BUILD_OUT" branch --list "$GH_PAGES_BRANCH")" ]; then
+        # create the destination branch if it doesn't exist
+        git -C "$BUILD_OUT" symbolic-ref HEAD "refs/heads/${GH_PAGES_BRANCH}"
+        rm -rf "${BUILD_OUT}/.git/index"
+        git -C "$BUILD_OUT" clean -fdx
+        git -C "$BUILD_OUT" checkout develop .gitignore
+        git -C "$BUILD_OUT" add .
+        git -C "$BUILD_OUT" commit --allow-empty -m "Initial commit for github pages"
+    else
```
cwillisf, line 209: Is this different only because a new `gh-pages` branch shouldn't contain the source code (and other contents) from `develop`? Or is there more going on here?

cwillisf (commented):
Looking good! I have some questions about potential simplification, but if there's a reason to keep it the way it is then I think that's fine :)

scripts/build-monorepo.sh
MiroslavDionisiev: The reason is that after we changed once the root to be treated as directory with `--to-subdirectory-filter` the rerun of the command with different values keeps the structure created by the first call. This may have a better solution but I decided to not try and over-engineer it.

scripts/build-monorepo.sh
MiroslavDionisiev: My intention was to create an orphan branch so that its history is not associated with the branch it is checked out of. However I didn't like the state of the branch after `git checkout --orphan` and decided to go with this solution.

scripts/build-monorepo.sh
```diff
@@ -462,6 +467,12 @@ process_workspace_webpack_config () {
     done
 }
 
+use_node_version_from_nvmrc () {
+    source ~/.nvm/nvm.sh
+    nvm install < .nvmrc
+    nvm use < .nvmrc
```
cwillisf: I don't think `nvm` uses `stdin` at all, so I believe these redirects are unnecessary. As long as `.nvmrc` is present in the current working directory, this should work the same:
```suggestion
    nvm install
    nvm use
```

scripts/git-batch-push.sh
```diff
@@ -0,0 +1,50 @@
...
+
+    # push each batch
+    for i in $(seq $n -$BATCH_SIZE 1); do
+        # get the hash of the commit to push
+        h=$(git -C "$BUILD_OUT" log --first-parent --reverse --format=format:%H --skip $i -n1)
+        echo "Pushing $h..."
+        git -C "$BUILD_OUT" push $REMOTE ${h}:refs/heads/$BRANCH --force
+    done
+    # push the final partial batch
+    git -C "$BUILD_OUT" push $REMOTE HEAD:refs/heads/$BRANCH --force
```
cwillisf: Is `--force` necessary? It makes me nervous about the possibility of something going wrong and causing this to force-push to the wrong branch or repo :sweat_smile:

If possible, I'd prefer to remove `--force` here and instead require a human to prepare the remote branch, maybe with a manual force-push to a base commit.

cwillisf (changes requested):

scripts/git-batch-push.sh
MiroslavDionisiev: Yeah, my idea was this value to be replaced with where the fork of the one running the script is, however I didn't make it obvious.

cwillisf (approved):

PR #8, opened 2024-08-26
[UEPR-33] Changed node version and updated import names

MiroslavDionisiev:
### Jira Ticket
[UEPR-33](https://scratchfoundation.atlassian.net/browse/UEPR-33)

### Proposed Changes
- updated node version of `scratch-editor`
- changed directory name from `workspaces` to `packages`
- added `@scracth` organization prefix to imports

[UEPR-33]: https://scratchfoundation.atlassian.net/browse/UEPR-33?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/build-monorepo.sh
```diff
@@ -383,11 +384,16 @@ fixup_branch () {
         for DEP in $PEERDEPS; do
             npm -C "$BUILD_OUT" install --force --save-peer --save-exact "$NPM_ORGANIZATION/$DEP" -w "$NPM_ORGANIZATION/$REPO" || package_replacement_error "$REPO" "$BRANCH" "$PEERDEPS"
         done
+
+        # replace the name of the package with the organozation prefixed one
```
cwillisf:
```suggestion
        # replace the name of the package with the organization prefixed one
```

scripts/build-monorepo.sh
```diff
@@ -354,22 +353,24 @@ fixup_branch () {
...
                 OPTDEPS="$OPTDEPS $DEP@*"
             fi
-            if jq -e .peerDependencies.\"$DEP\" "${BUILD_OUT}/workspaces/${REPO}/package.json" > /dev/null; then
-                jq  "del(.peerDependencies.\"$DEP\")" "${BUILD_OUT}/workspaces/${REPO}/package.json"  | sponge "${BUILD_OUT}/workspaces/${REPO}/package.json"
+            if jq -e .peerDependencies.\"$DEP\" "${BUILD_OUT}/packages/${REPO}/package.json" > /dev/null; then
+                jq  "del(.peerDependencies.\"$DEP\")" "${BUILD_OUT}/packages/${REPO}/package.json"  | sponge "${BUILD_OUT}/packages/${REPO}/package.json"
                 PEERDEPS="$PEERDEPS $DEP@*"
             fi
+
+            npm -C "$BUILD_OUT" uninstall "$DEP"
```
cwillisf, line 373: Could you explain what this line is here for?

cwillisf (commented):
Seems reasonable. I wonder how I missed the `packages/*` suggestion in the NPM docs. I'm pretty sure I started this after npm/cli#4426 but oh well :)

Anyway, one typo and I'd like to understand the need for the uninstall bit, but it seems pretty much :+1:

scripts/build-monorepo.sh
MiroslavDionisiev: In the previous PR, I forgot to update the import names of packages because the applications were building properly. The reason was in node_modules there were both packages with and without the '@scratch' prefix and the imports were finding what they were pointing to. This way we can verify that after we clear the node_modules from the old ones, the regex replaces the package name at all necessary places.

cwillisf: Ah, that makes sense. Thanks!

cwillisf (approved):

PR #19, opened 2024-09-10
Gha integration

MiroslavDionisiev:


scripts/build-packages.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+set -e
+
+cd ./packages/scratch-svg-renderer
+npm run build
+cd -
```
cwillisf: No change necessary, and I'm not sure why I didn't think of this last time around, but this could be further simplified to:
```suggestion
npm -C ./packages/scratch-svg-renderer run build
```

cwillisf (approved):

PR #25, opened 2024-09-17
Provide a standalone build of GUI and configure TS support

georgyangelov:
- Added TS support so that and configuration interfaces we write can have exported types.
- Removed react, react-dom, redux and react-redux from the regular bundle (`scratch-gui.js`) as they weren't used by scratch-www anyway
- Added a separate output bundle called scratch-gui-standalone which includes all of the dependencies above and exposes a function to instantiate the editor with its own React instance. This allows scratch-web and scratch-gui to not conflict on React versions - one can use 16, the other can use 18 with no issues.

I haven't tested the build with scratch-www @MiroslavDionisiev what's the simplest way to do that? Want to make sure the externals were not actually used from scratch-www.

---

I will make some changes to abstract away project loading soon, but that would probably be a separate PR


georgyangelov:
The build fails as this PR depends on changes from https://github.com/scratchfoundation/scratch-webpack-configuration/pull/54 which are not in NPM yet

packages/scratch-gui/webpack.config.js
```diff
@@ -13,16 +13,25 @@ const baseConfig = new ScratchWebpackConfigBuilder(
...
     })
     .setTarget('browserslist')
     .merge({
         output: {
             assetModuleFilename: 'static/assets/[name].[hash][ext][query]',
+            // Forces lookup for assets at the root of the current origin. Otherwise
+            // it is resolved dynamically based on what bundles the library next.
+            // For example, if scratch-gui is later bundled to `/js/`, the assets will
+            // be looked up at `/js/static/assets/...`, which is incorrect (for Scratch).
+            publicPath: '/',
```
georgyangelov: @cwillisf This should fix the `/js/static/assets` issue - once it's integrated into `scratch-www` we'll need to edit the copy plugin settings again.

Props to @MiroslavDionisiev for finding and testing this config change

packages/scratch-gui/src/index-standalone.tsx
```diff
@@ -0,0 +1,51 @@
...
+            guiMiddleware,
+        ),
+    );
+
+    return {
+        render(props: GUIProps) {
+            ReactDOM.render(
+                <Provider store={store}>
+                    <GUI {...props} />
+                </Provider>,
```
MiroslavDionisiev, line 49: Shouldn't we wrap `GUI` in `IntlProvider`?

MiroslavDionisiev (approved):

packages/scratch-gui/src/index-standalone.tsx
georgyangelov: I don't think we need to, seems translations work fine:

Screenshot 2024-09-18 at 13 54 29

packages/scratch-gui/tsconfig.json
```diff
@@ -0,0 +1,111 @@
+{
+  "include": ["src"],
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig to read more about this file */
+
+    /* Projects */
+    // "incremental": true,                              /* Save .tsbuildinfo files to allow for incremental compilation of projects. */
+    // "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
```
cwillisf: Are all these commented lines here to show the defaults?

packages/scratch-gui/webpack.config.js
cwillisf: Should this be an override-able default in `scratch-webpack-configuration`?

cwillisf (approved):

packages/scratch-gui/tsconfig.json
georgyangelov: Actually not, this is the auto-generated file where non-commented options are either required or changed. The commented ones are not the defaults (weird, yes). It's probably best to remove the commented ones now that I think about it, what do you think?

packages/scratch-gui/webpack.config.js
georgyangelov: Good question, I didn't want to make a defaults change in `scratch-webpack-configuration` since I'm not sure where else it's used, but if you think that makes sense (which I agree) I'll move it there

cwillisf: Yeah, fair point.

To answer your implied question: this config is currently being used by `scratch-gui`, `scratch-render`, `scratch-svg-renderer`, and `scratch-vm`. Since I'm blocked on my "main" project right now, I think I can spend some time today checking on whether `publicPath: '/'` would be good, bad, or neutral for each of those.

Really, the phrase "resolved dynamically" caused an allergic reaction for me ;)

packages/scratch-gui/tsconfig.json
cwillisf: Yeah, I guess I'd prefer to have them removed, unless you can think of a reason to keep them. If it's just to softly enforce the order of options in this file, I think it's OK to remove :)

packages/scratch-gui/webpack.config.js
cwillisf: OK, I tried `publicpath: '/'` in all four of those repositories. There was no effect in any of them except `scratch-gui`. I experimented with other values of `publicPath`, including `'.'`, and my conclusion is that `'/'` makes sense as a default but it does need to allow for override.

For the record, here's what happens in a JS bundle when you change `publicPath` from unset to `/`:
```diff
 /******/       /* webpack/runtime/publicPath */
 /******/       (() => {
-/******/               var scriptUrl;
-/******/               if (__webpack_require__.g.importScripts) scriptUrl = __webpack_require__.g.location + "";
-/******/               var document = __webpack_require__.g.document;
-/******/               if (!scriptUrl && document) {
-/******/                       if (document.currentScript && document.currentScript.tagName.toUpperCase() === 'SCRIPT')
-/******/                               scriptUrl = document.currentScript.src;
-/******/                       if (!scriptUrl) {
-/******/                               var scripts = document.getElementsByTagName("script");
-/******/                               if(scripts.length) {
-/******/                                       var i = scripts.length - 1;
-/******/                                       while (i > -1 && (!scriptUrl || !/^http(s?):/.test(scriptUrl))) scriptUrl = scripts[i--].src;
-/******/                               }
-/******/                       }
-/******/               }
-/******/               // When supporting browsers where an automatic publicPath is not supported you must specify an output.publicPath manually via configuration
-/******/               // or pass an empty string ("") and set the __webpack_public_path__ variable from your code to use your own logic.
-/******/               if (!scriptUrl) throw new Error("Automatic publicPath is not supported in this browser");
-/******/               scriptUrl = scriptUrl.replace(/#.*$/, "").replace(/\?.*$/, "").replace(/\/[^\/]+$/, "/");
-/******/               __webpack_require__.p = scriptUrl;
+/******/               __webpack_require__.p = "/";
 /******/       })();
```

georgyangelov: I made the change in the [scratch-webpack-configuration](https://github.com/scratchfoundation/scratch-webpack-configuration/pull/54) PR

packages/scratch-gui/tsconfig.json
georgyangelov: Removed the extra comments

georgyangelov:
I realized I hadn't configured eslint to check TS files and corrected that.

@cwillisf does everything in scratch-webpack-configuration and here looks ok to you? Do we push the buttons? 😄

cwillisf (approved):
Looks good! Merge (both) when ready :)

PR #29, opened 2024-10-02
Feature/uepr 65 confirmation prompt when deleting

KManolov3:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-65

### Proposed Changes

Adds a prompt modal when attempting to delete an asset. 

image

### Test Coverage

Updated corresponding unit and integration tests
Added a new unit test for the cases that include a confirmation modal


packages/scratch-gui/test/unit/containers/sprite-selector-item.test.jsx
```diff
@@ -48,11 +49,21 @@ describe('SpriteSelectorItem Container', () => {
         onDeleteButtonClick = jest.fn();
         dispatchSetHoveredSprite = jest.fn();
         selected = true;
+        vm = new VM();
+        store = mockStore({scratchGui: {
+            hoveredTarget: {receivedBlocks: false, sprite: null},
+            assetDrag: {dragging: false},
+            vm
+        }});
```
cwillisf, line 58: Thanks for making the test work even though you had to jump through extra hoops!

cwillisf (changes requested):
The code looks good to me, but this has made me realize that we didn't include a UX design with this ticket. Sorry about that! We should talk to Tif about the design for this dialog.

packages/scratch-gui/src/components/delete-confirmation-prompt/delete-confirmation-prompt.jsx
```diff
@@ -0,0 +1,185 @@
...
+    onCancel,
+    onOk,
+    modalPosition,
+    entityType,
+    relativeElemRef
+}) => {
+    const modalPositionValues = calculateModalPosition(relativeElemRef, modalPosition);
+
+    return (<ReactModal
+        isOpen
```
MiroslavDionisiev, line 93: From where do we get `isOpen`?

packages/scratch-gui/src/components/delete-confirmation-prompt/delete-confirmation-prompt.jsx
```diff
@@ -0,0 +1,185 @@
...
+                bottom: 0,
+                zIndex: 510,
+                backgroundColor: 'transparent'
+            }
+        }}
+        contentLabel={intl.formatMessage(messages.confirmDeletionHeading)}
+        onRequestClose={onCancel}
+    >
+        <Box className={styles.modalContainer}>
+            { modalPosition === 'right' ?
```
MiroslavDionisiev, line 123: We can avoid using the ternary operator in this case.

packages/scratch-gui/src/components/delete-confirmation-prompt/delete-confirmation-prompt.jsx
```diff
@@ -0,0 +1,185 @@
...
+                            className={styles.deleteIcon}
+                            src={undoIcon}
+                        />
+                        <div className={styles.message}>
+                            <FormattedMessage {...messages.cancelOption} />
+                        </div>
+                    </button>
+                </Box>
+            </Box>
+            {modalPosition === 'left' ?
```
MiroslavDionisiev, line 163: ditto

packages/scratch-gui/src/containers/sprite-selector-item.jsx
```diff
@@ -126,27 +144,36 @@ class SpriteSelectorItem extends React.PureComponent {
             receivedBlocks,
             costumeURL,
             vm,
+            deleteConfirmationModalPosition,
             /* eslint-enable no-unused-vars */
             ...props
         } = this.props;
-        return (
+        return (<>
+            {this.state.isDeletePromptOpen ? <DeleteConfirmationPrompt
```
MiroslavDionisiev, line 152: ditto

MiroslavDionisiev:
The text colour of the buttons seems black

packages/scratch-gui/src/components/delete-confirmation-prompt/delete-confirmation-prompt.jsx
KManolov3: This prop means it's always open (when rendered). We practically create a new instance of the modal component each time we want to show it, with different data. Chose to implement it this way to be consistent with other modals throughout the app.

packages/scratch-gui/src/components/delete-confirmation-prompt/delete-confirmation-prompt.jsx
KManolov3: What do you propose instead of a ternary?

KManolov3: {cond && <Component />} syntax?

MiroslavDionisiev: Yeah we can do `condition && component`. That was my intention, forgot to write it down, sorry

KManolov3: You are correct, as long as `condition` is a boolean. If it is something else that evaluates to another truthy value - string, number -> React will render it and cause unexpected behaviour/error. Since it's easy to make such a change (e.g. refactor `myNum > 0 &&` to just `myNum &&`) without noticing, I prefer to consistently use ternaries with `null` in the else case.

MiroslavDionisiev (approved):

PR #30, opened 2024-10-09
[UEPR-44] exported redux action and added query param showing what tutorial is opened

MiroslavDionisiev:
### Jira Ticket

[UEPR-44](https://scratchfoundation.atlassian.net/browse/UEPR-44)

### Proposed Changes
Exports redux action for selection a tutorial, shows tutorials query parameter that was missing

[UEPR-44]: https://scratchfoundation.atlassian.net/browse/UEPR-44?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/scratch-gui/src/containers/tips-library.jsx
```diff
@@ -58,6 +58,7 @@ class TipsLibrary extends React.PureComponent {
             return window.open(window.location.origin + urlParams, '_blank');
         }
 
+        history.pushState({}, {}, `?tutorial=${item.urlId}`);
```
georgyangelov: Why do we need these?

MiroslavDionisiev: I added it mainly for consistency. For the tutorials that require to be viewed with a specific project as in the if case above we add the query parameter. Also when we are opening tutorials page from ideas page first we have the query parameter and after the new project is created and we change the url the parameters are omitted. Then I decided to use the parameter to know if the user has opened the editor through the tutorials and not trigger the pop up

georgyangelov (approved):

MiroslavDionisiev:

MiroslavDionisiev:
recheck

PR #32, opened 2024-10-10
Allow externally-configurable storage in scratch-gui

georgyangelov:
This PR enables scratch-gui to receve a GUIConfig object externally. This object currently allows for configuration of the ScratchStorage instance, plus the project save function.

I've made it backwards-compatible in that there is a default implementation of GUIConfig that is used for the non-standalone version which should behave the same way as previously.

The config is stored in Redux on initial load and is used from there by the higher-order components that do data loading/saving. Most of the changes here are to make the ScratchStorage instance dynamic as opposed to a globally-accessible singleton as it way before.

See https://github.com/scratchfoundation/scratch-platform/pull/60 for the other side of the config.


georgyangelov:
@cwillisf would love to hear your thoughts on the general approach 🙏

cwillisf:
I haven't read the whole thing yet; I just tried to get an overview. So far, I like where you're heading!

One thought: some groups using scratch-gui as a library might not be using TS, so they might want to supply GUI configuration as a plain JS object. Obviously, that use case isn't currently supported since there was no such thing as a GUI config object until now! But if it's not too difficult to support that, I think it would be nice to have & retain that option. If it becomes a pain or gets in the way of good types, I think it's OK for this new feature to require TS.

georgyangelov:
Thank you!

> One thought: some groups using scratch-gui as a library might not be using TS, so they might want to supply GUI configuration as a plain JS object. Obviously, that use case isn't currently supported since there was no such thing as a GUI config object until now! But if it's not too difficult to support that, I think it would be nice to have & retain that option.

That is possible now as well. The `GUIConfig`/`GUIStorage` types are just interfaces so it ends up deleted after compilation and it's all pure JS from that point on. So JS users can pass a simple JS object with the same keys and it will work. Of course, if they provide the wrong keys or types -> then it's runtime errors, but that's a given.

georgyangelov:
@cwillisf this should be ready to get a further look at as well. Only thing left is to update the new scratch-storage version once that is in.

cwillisf (approved):
:sunglasses:

PR #36, opened 2024-10-28
feat: uepr-38: implement debugging modal with text topics

KManolov3:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-38

### Proposed Changes

Implements a debugging modal with text topics

### Reason for Changes

### Test Coverage


packages/scratch-gui/webpack.config.js
```diff
@@ -132,7 +132,16 @@ const buildConfig = baseConfig.clone()
...
+        title: 'Scratch 3.0 GUI',
+        // Google Tag Manager ID
+        // Looks like 'GTM-XXXXXXX'
+        gtm_id: process.env.GTM_ID || '',
+
+        // Google Tag Manager env & auth info for alterative GTM environments
+        // Looks like '&gtm_auth=0123456789abcdefghijklm&gtm_preview=env-00&gtm_cookies_win=x'
+        // Taken from the middle of: GTM -> Admin -> Environments -> (environment) -> Get Snippet
+        // Blank for production
+        gtm_env_auth: process.env.GTM_ENV_AUTH || ''
```
cwillisf: For the sake of consistency, I'd recommend moving this stuff into something like:
```js
const commonHtmlWebpackPluginOptions = {
    // Google Tag Manager ID
    gtm_id: // etc
};
```

Then you can add something like this here and in all the other pages:
```js
    .addPlugin(new HtmlWebpackPlugin({
        ...commonHtmlWebpackPluginOptions,
        chunks: // etc
```

That way, all of the generated pages have the same GTM config without needing to manually copy it to each one.

cwillisf: It's also worth checking with Georgi regarding how this might affect GUI stand-alone mode, if you haven't already, but I imagine it's probably fine.

MiroslavDionisiev (approved):

packages/scratch-gui/webpack.config.js
KManolov3: Extracted the config in a common variable and reused it for the configuration of the rest of the `HtmlWebpackPlugin`'s. Was that what you had in mind?

On your second point - yes, I coordinated this with Georgi.

PR #53, opened 2025-01-13
Allow state and rendering to be managed separately in the standalone version

georgyangelov:
### Reason for Changes

Helps with NGP integration - this way when we transition from page to page (e.g. from the project page to the "See inside" one) we don't need to reinitialize everything.

### Proposed Changes

- Introduce an `EditorState` class which wraps and creates the editor Redux state (extracted from AppStateHOC). It's essentially the non-rendering parts of `AppStateHOC`.
- Make AppStateHOC create and use the EditorState class internally
- Add an AppStateProviderHOC which is the same as AppStateHOC except it accepts the EditorState from above and doesn't create it by itself
- For an example of how this is used in NGP, see packages/scratch-web/src/components/editor/editor-state-provider.tsx and https://github.com/scratchfoundation/scratch-platform/pull/137/files in https://github.com/scratchfoundation/scratch-platform/pull/137/files

This contains breaking changes only to the standalone version - the non-standalone bundle as currently used by scratch-www should not be affected (although I haven't tested it explicitly).

I also added `referrerPolicy="no-referrer"` to the avatar rendering - that makes an issue loading Google avatars go away (they seem to do rate-limiting based on a referer if it's present). IMO this shouldn't hurt to have it for non-Google linked avatars as well, but let me know if you think otherwise.


georgyangelov:

georgyangelov:
Looks like I have some test issues, will take a look tomorrow

cwillisf (approved):
Looks good to me.

When I read the description, I initially thought, "Why not put this state in Redux to separate it from rendering?" But then I started reading the code, and I [understood](https://media4.giphy.com/media/v1.Y2lkPTc5MGI3NjExd2E3bmF4OTNvYWs4eWw4dGlyeTlzeGM4bG5hdHNwYmI1dW81end1bSZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/orVa44Oav5WoF1LVOE/giphy.gif).

If you can think of a reasonable one, it would be cool to add a test that would have failed before this PR but would succeed now. I'm not stressed about it, though.

It still makes me smile to see TS action in `scratch-gui` :)

MiroslavDionisiev (approved):

cwillisf (approved):
LGTM! Thanks for adding tests.

PR #56, opened 2025-01-23
fix(gui typing): make params optional

YoanBabulkov:
### Proposed Changes
Changes the fields in `params` object argument inside `saveProject` func to be optional 

### Reason for Changes
Linked to NGP - when we are not remixing, we receive an empty `params` object (so the fields are actually optional) and we need to reflect that in `GUIStorage` so that we have correct typing


YoanBabulkov:

georgyangelov (approved):

PR #57, opened 2025-01-24
fix: missing dist outputs in published packages

MiroslavDionisiev:
### Resolves

Executing `npm publish --access=public --workspace={package-name}` packs and publishes a package in the context of a workspace but it doesn't ignore only the files in the `.npmignore` in the workspace but looks into the root as well. Because `.npmignore` file is missing in the root of the editor npm uses the `.gitignore` one. This leads to publishing wrong files and directories.
In addition the standalone repositories publish unnecessary content for the functioning of the package. Setting `files` in package.jsons allows to list what we want to be included in the published package. Current packages will contain only:
- dist output
- package.json
- README
- LICENSE

### DIFF
--- a/packages/scratch-gui/package.json
+++ b/packages/scratch-gui/package.json
@@ -21,6 +21,7 @@
       "default": "./dist/scratch-gui-standalone.js"
     }
   },
+  "files": ["dist"],
   "scripts": {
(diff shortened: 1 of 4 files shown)


cwillisf (approved):
I'm not sure what to do about it yet, but it seems like a published package should not have `exports` entries that refer to files that are excluded from its published contents. This change makes the package less broken, though, so 👍

cwillisf (approved):

PR #68, opened 2025-01-27
Feature/uepr 140 merge develop in scratch desktop

KManolov3:
### Resolves
Part of https://scratchfoundation.atlassian.net/browse/UEPR-140

### Proposed Changes
Integrate newest scratch-gui changes into scratch-desktop branch.

### Reason for Changes
This is one of the necessary steps to catch up scratch-android and scratch-desktop with functionality released in the last months. The end goal will be to merge all three branches into one, which will make easier future desktop and android releases.


packages/scratch-gui/package.json
```diff
@@ -16,6 +16,18 @@
       "types": "./dist/types/index.d.ts",
       "default": "./dist/scratch-gui.js"
     },
+    "./costumes": {
```
KManolov3, line 19: This was due to the need to import these in scratch-desktop. We can't do it from the default import since we are using esModules, while the fetch script is with cjs. We _might_ be able to do it with another entrypoint which packages all the libraries together, but it seems like a quicker solution to do it like this - although accepting feedback on that point.

cwillisf: :melting_face: I wonder how much CJS vs ESM confusion has cost the industry as a whole

This seems fine to me for now. We're talking about refactoring the media libraries anyway, and based on recent conversations, it might even happen this year. When we do that, I suspect `scratch-desktop` (or whatever form it's in at that future time) will grab the libraries a different way, and these exports will be removed. If that doesn't happen... this isn't terrible.

packages/scratch-gui/src/components/library-item/library-item.jsx
```diff
@@ -36,10 +37,12 @@ class LibraryItemComponent extends React.PureComponent {
...
-                    <img
-                        className={styles.featuredImage}
-                        src={this.props.iconURL}
-                    />
+                    {this.props.iconSource ? (
+                        <ScratchImage
+                            className={styles.featuredImage}
+                            imageSource={this.props.iconSource}
+                        />
+                    ) : null}
```
cwillisf, line 45: Future: I'd like to add general download-queue functionality somewhere central, maybe `scratch-storage`, and funnel these loads through that. Doing so should simultaneously obviate `ScratchImage` here _and_ make it so we can reliably load projects with thousands of assets without the browser freaking out. :muscle:

No action necessary right now, just sharing thoughts...

packages/scratch-gui/webpack.config.js
```diff
@@ -99,6 +99,11 @@ const distConfig = baseConfig.clone()
             'scratch-gui': path.join(__dirname, 'src/index.ts')
         },
         output: {
+            // TODO: Test if this breaks the current version of -www!!
```
cwillisf, line 102: @MarshTheTester, something to look out for once we merge this PR. If this is a problem, it'll likely break everything, but it's still worth looking out for subtle problems in the music extension or other non-obvious places.

cwillisf (approved):

PR #138, opened 2025-02-05
fix: correct 'this' context when calling `saveProjectThumbnail`

YoanBabulkov:
### Proposed Changes

Wrap the `storage.saveProject` function in an arrow function when passing to `onUpdateProjectThumbnail`

### Reason for Changes
We need to preserve the correct `this` binding when invoking the `saveProjectThumbnail`. Wrapping in an arrow function ensures we invoke the function from the correct object (storage)

### DIFF
--- a/packages/scratch-gui/src/lib/project-saver-hoc.jsx
+++ b/packages/scratch-gui/src/lib/project-saver-hoc.jsx
@@ -427,8 +427,9 @@ const ProjectSaverHOC = function (WrappedComponent) {
             isManualUpdating: getIsManualUpdating(loadingState),
             loadingState: loadingState,
             locale: state.locales.locale,
-            onUpdateProjectThumbnail: ownProps.onUpdateProjectThumbnail ??
-                ((projectId, thumbnail) => storage.saveProjectThumbnail(projectId, thumbnail)),
+            onUpdateProjectThumbnail:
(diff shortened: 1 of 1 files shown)


georgyangelov (approved):

packages/scratch-gui/src/lib/project-saver-hoc.jsx
```diff
@@ -427,7 +427,8 @@ const ProjectSaverHOC = function (WrappedComponent) {
             isManualUpdating: getIsManualUpdating(loadingState),
             loadingState: loadingState,
             locale: state.locales.locale,
-            onUpdateProjectThumbnail: ownProps.onUpdateProjectThumbnail ?? storage.saveProjectThumbnail,
+            onUpdateProjectThumbnail: ownProps.onUpdateProjectThumbnail ??
+                ((projectId, thumbnail) => storage.saveProjectThumbnail(projectId, thumbnail)),
```
georgyangelov: Previously this was `undefined` if the storage didn't have the `storage.saveProjectThumbnail` function. Now it's a function that will throw an error because `storage.saveProjectThumbnail` doesn't exist. Please handle that case.

georgyangelov (changes requested):

PR #139, opened 2025-02-06
feat: [UEPR-56] Migrate to React 18

Bogomil-Stoyanov:
**Ticket:**
[UEPR-56](https://scratchfoundation.atlassian.net/browse/UEPR-56?atlOrigin=eyJpIjoiOGIxNzhiZDVmMjY4NDRmYjhlZmM3ZGExN2RlMzU5ODQiLCJwIjoiaiJ9)

This pull request includes an upgrade to the intl library and react 18.

[UEPR-56]: https://scratchfoundation.atlassian.net/browse/UEPR-56?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/scratch-gui/test/unit/components/icon-button.test.jsx
```diff
@@ -1,37 +1,56 @@
...
+        if (!button) {
+            throw new Error('Button not found');
+        }
+
+        fireEvent.click(button);
+        
+        expect(handleClick).toHaveBeenCalledTimes(1);
+    });
+
+    test('matches snapshot', () => {
```
MiroslavDionisiev: If the component matches the snapshot it would be the same as the result of `renders with all props correctly`

packages/scratch-gui/test/unit/components/button.test.jsx
```diff
@@ -1,23 +1,27 @@
...
 import ButtonComponent from '../../../src/components/button/button';
-import renderer from 'react-test-renderer';
 
 describe('ButtonComponent', () => {
     test('matches snapshot', () => {
-        const onClick = jest.fn();
-        const component = renderer.create(
-            <ButtonComponent onClick={onClick} />
+        const { container } = render(
+            <ButtonComponent onClick={jest.fn()} />
```
MiroslavDionisiev, line 9: At some places it is used `const onClick = jest.fn();` and at others `jest.fn()` is passed directly to onClick. We should unify the syntax everywhere

packages/scratch-gui/test/unit/components/button.test.jsx
```diff
@@ -1,23 +1,27 @@
...
         );
-        expect(component.toJSON()).toMatchSnapshot();
+        
+        expect(container.firstChild).toMatchSnapshot();
     });
 
     test('triggers callback when clicked', () => {
```
MiroslavDionisiev, line 15: I would add a test that the callback is not triggered if not clicked.

packages/scratch-gui/test/unit/components/cards.test.jsx
```diff
@@ -36,27 +37,31 @@ describe('Cards component', () => {
...
+    
+        const imageStep = container.querySelector('img');
+        expect(imageStep).toBeTruthy();
 
-        const imageStep = component.find(ImageStep);
-        expect(imageStep.props().image).toEqual('id1 - img');
-        expect(imageStep.props().title).toEqual('id1 - name');
+        const divWithText = [...container.querySelectorAll('div')].find(el => el.textContent.includes('id1 - name'));
+        expect(divWithText).toBeTruthy();
+        expect(divWithText).toHaveTextContent('id1 - name');
```
MiroslavDionisiev: if `divWithText` is truthy wouldn't that mean it has the text context, because we use it in the query selector. Same for the image

packages/scratch-gui/test/unit/components/icon-button.test.jsx
```diff
@@ -1,37 +1,56 @@
...
-    test('triggers callback when clicked', () => {
-        const onClick = jest.fn();
-        const title = <div>Text</div>;
-        const imgSrc = 'imgSrc';
-        const componentShallowWrapper = shallow(
-            <IconButton
-                img={imgSrc}
-                title={title}
-                onClick={onClick}
+    test('triggers callback only once when clicked', () => {
```
MiroslavDionisiev: I would add a test that the callback is not triggered if not clicked.

packages/scratch-gui/test/unit/components/icon-button.test.jsx
```diff
@@ -1,37 +1,56 @@
...
             />
         );
-        componentShallowWrapper.simulate('click');
-        expect(onClick).toHaveBeenCalled();
+        
+        const button = container.querySelector('div[role="button"]');
+        
+        if (!button) {
+            throw new Error('Button not found');
+        }
```
MiroslavDionisiev: I am not sure if we need such error handling here

packages/scratch-gui/test/unit/components/icon-button.test.jsx
MiroslavDionisiev: I would rather leave the snapshot test rather than `renders with all props correctly`. The second one will not fail if we add new props to the component and we may miss updating the test. Whereas if the snapshot is old it will fail

Bogomil-Stoyanov (commented):
🤝

jnnck:
Is there a release target on this one? We've been waiting on a react update for years now... Seems like you guys are close?

KManolov3:
> Is there a release target on this one? We've been waiting on a react update for years now... Seems like you guys are close?

This was frozen for a while due to higher priority work, but now there is a fair chance we release this soon.

packages/scratch-gui/src/components/context-menu/context-menu.css
```diff
@@ -2,37 +2,42 @@
...
 @import "../../css/z-index.css";
 
-.context-menu {
-    min-width: 130px;
-    padding: 5px 0; /* The white strip at the top and bottom of the menu */
-    margin: 2px 0 0; /* To keep the menu below the cursor comfortably */
-    font-size: 0.85rem;
-    text-align: left;
-    background-color: $ui-white;
+.context-menu-content {
```
cwillisf, line 5: Inconsistent indenting in this file

packages/scratch-gui/src/components/webgl-modal/webgl-modal.jsx
```diff
@@ -14,88 +14,96 @@ const messages = defineMessages({
...
+                            defaultMessage="To learn more, go to the <a>{previewFaqLink}</a>."
+                            description="Scratch 3.0 FAQ description"
+                            id="gui.webglModal.previewfaq"
+                            values={{
+                                previewFaqLink: (
                                     <FormattedMessage
                                         defaultMessage="FAQ"
                                         description="link to Scratch 3.0 FAQ page"
-                                        id="gui.webglModal.previewfaqlinktext"
+                                        id="gui.unsupportedBrowser.previewfaqlinktext"
```
cwillisf: Was this change intentional?

packages/scratch-gui/src/containers/sprite-selector-item.jsx
```diff
@@ -127,8 +127,10 @@ class SpriteSelectorItem extends React.PureComponent {
         this.setState({isDeletePromptOpen: false});
     }
     setRef (component) {
-        // Access the DOM node using .elem because it is going through ContextMenuTrigger
-        this.ref = component && component.elem;
+        // The ref is on the element inside ContextMenu.Trigger
+        if (component) {
+            this.ref = component;
+        }
```
cwillisf: Should `this.ref` be set to `null` (or something) if `component` is falsy?

packages/scratch-gui/src/lib/radix-ui-context-menu.js
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable import/no-unresolved */
+/*
+    https://github.com/import-js/eslint-plugin-import/issues/1810
+    eslint-plugin-import is not aware of exports definition in package.json
+    meaning we should disable linting for this import or use require instead
+    moved the import in a separate file so that the disabling happens in one place only
+*/
```
cwillisf, line 7: Thanks for the explanation!

packages/scratch-gui/test/unit/components/button.test.jsx
cwillisf: I would standardize on `const onClick = jest.fn();` and call `expect(onClick).toHaveBeenCalledTimes` with the expected number of clicks (probably 0 or 1). If there's a specific reason not to check, like if it's truly irrelevant to the test at hand, then :iwasneverhere:.

packages/scratch-gui/test/unit/components/controls.test.jsx
```diff
@@ -14,27 +13,36 @@ describe('Controls component', () => {
...
-        component.find(GreenFlag).simulate('click');
-        expect(props.onGreenFlagClick).toHaveBeenCalled();
-
-        component.find(StopAll).simulate('click');
-        expect(props.onStopAllClick).toHaveBeenCalled();
+    test('does not show turbo mode when not in turbo mode', () => {
+        const { container: containerNoTurbo } = renderWithIntl(<Controls {...defaultProps()} />);
+        const noTurboMode = [...containerNoTurbo.querySelectorAll('div')].reverse().find(el => el.textContent.includes('Turbo Mode'));
+        expect(noTurboMode).toBeFalsy();
     });
```
cwillisf, line 25: 😎

packages/scratch-gui/test/unit/components/controls.test.jsx
```diff
@@ -14,27 +13,36 @@ describe('Controls component', () => {
...
-});
+
+    describe('triggers the right callbacks when clicked', () => {
+        test('when green flag button clicked tirggers the right callback', () => {
```
cwillisf: Typo: `tirggers` (copied below as well)

packages/scratch-gui/test/unit/containers/sound-editor.test.jsx
```diff
@@ -31,250 +40,243 @@ describe('Sound Editor Container', () => {
...
-                soundIndex={soundIndex}
-                store={store}
-            />
-        );
-        const component = wrapper.find(SoundEditorComponent);
-        component.props().onEcho();
-        expect(mockAudioEffects.instance.name).toEqual(mockAudioEffects.effectTypes.ECHO);
-        expect(mockAudioEffects.instance.process).toHaveBeenCalled();
-    });
-
```
cwillisf, line 191: Why was this removed instead of following the same pattern as the others?

packages/scratch-gui/package.json
```diff
@@ -46,7 +46,7 @@
     "i18n:push": "tx-push-src scratch-editor interface translations/en.json",
     "i18n:src": "rimraf ./translations/messages/src && babel src > tmp.js && rimraf tmp.js && build-i18n-src ./translations/messages/src ./translations/",
     "start": "webpack serve",
-    "test": "npm run test:lint && npm run test:unit && npm run test:integration",
+    "test": "npm run test:lint && npm run test:unit && npm run build && npm run test:integration",
```
cwillisf: I'd prefer not to build as part of `npm run test` since `.github/workflows/ci.yml` builds in one job before it tests in a completely separate job.

cwillisf (changes requested):
Wow, this is an amazing amount of work. Thank you so much for this effort!

packages/scratch-gui/test/unit/components/button.test.jsx
MiroslavDionisiev: In this case it may be irrelevant as we compare the component to a snapshot

packages/scratch-gui/package.json
MiroslavDionisiev: Yeah, probably some bad merge conflicts resolving on my part

packages/scratch-gui/test/unit/containers/sound-editor.test.jsx
MiroslavDionisiev: The `Echo` button has been removed. The old tests were accessing the props: the `onEcho` callback which is still passed, and executing it and that's why the test was passing. While with the new testing framework we are searching for the button and clicking it.

packages/scratch-gui/src/containers/sprite-selector-item.jsx
MiroslavDionisiev: The fix isn't any good as it doesn't take into consideration the unmounting call and strange enough after reverting the change the issue that we faced is not getting reproduced. It may be a change in the context menu library we are using.

KManolov3:
Closing this PR, will push a branch into this remote and open a PR from it - the GH releases work only with branches in the same remote, so this will reduce the added overhead.

PR #172, opened 2025-02-18
feat: hide the tutorials that require a project

fouksf:
Add a new property to the editor that controls whether or not the tutorials that require an existing project should be shown.

### DIFF
--- a/packages/scratch-gui/src/containers/tips-library.jsx
+++ b/packages/scratch-gui/src/containers/tips-library.jsx
@@ -63,15 +63,24 @@ class TipsLibrary extends React.PureComponent {
     render () {
         const decksLibraryThumbnailData = Object.keys(decksLibraryContent)
             .filter(id => {
-                if (!isScratchDesktop() && !this.props.hideTutorialProjects) {
-                    return true;
-                }
+                /**
+                 * Scratch desktop can't support project and video-only tutorials.
+                 * NGP can't support project tutorials.
+                 * The online editor, conversely, should show all tutorials.
+                 */
+                
                 const deck = decksLibraryContent[id];
-                // Don't show tutorials that require specific projects
-                if (Object.prototype.hasOwnProperty.call(deck, 'requiredProjectId')) return false;
-                // Scratch Desktop should not load tutorials that are _only_ videos
-                if (isScratchDesktop() && deck.steps.filter(s => s.title).length === 0) return false;
-                // Allow any other tutorials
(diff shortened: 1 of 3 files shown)


fouksf:

Bogomil-Stoyanov (approved):

packages/scratch-gui/src/containers/tips-library.jsx
```diff
@@ -63,12 +63,14 @@ class TipsLibrary extends React.PureComponent {
     render () {
         const decksLibraryThumbnailData = Object.keys(decksLibraryContent)
             .filter(id => {
-                if (notScratchDesktop()) return true; // Do not filter anything in online editor
+                if (!isScratchDesktop() && !this.props.hideTutorialProjects) {
```
KManolov3: What do you think of trying to reduce the complexity a bit, for example:

````
.filter(id => {
                /**
                 * Scratch desktop can't support project and video-only tutorials.
                 * NGP can't support project tutorials.
                 * The online editor, conversely, should show all tutorials.
                 */

                const isProjectTutorial = Object.prototype.hasOwnProperty.call(deck, 'requiredProjectId');
                const isVideoOnlyTutorial = decksLibraryContent[id].steps.filter(s => s.title).length === 0;

                if (isProjectTutorial && (this.props.hideTutorialProjects || isScratchDesktop())) {
                    return false;
                }

                if (isVideoOnlyTutorial && isScratchDesktop()) {
                    return false;
                }

                return true;
            })
````

File: packages/scratch-gui/src/containers/tips-library.jsx [hunk 2]
```diff
@@ -63,13 +63,24 @@ class TipsLibrary extends React.PureComponent {
...
+                 */
+                
                 const deck = decksLibraryContent[id];
-                // Scratch Desktop doesn't want tutorials with `requiredProjectId`
-                if (Object.prototype.hasOwnProperty.call(deck, 'requiredProjectId')) return false;
-                // Scratch Desktop should not load tutorials that are _only_ videos
-                if (deck.steps.filter(s => s.title).length === 0) return false;
-                // Allow any other tutorials
+                const isProjectTutorial = Object.prototype.hasOwnProperty.call(deck, 'requiredProjectId');
+                const isVideoOnlyTutorial = decksLibraryContent[id].steps.filter(s => s.title).length === 0;
```
KManolov3, line 74: I thought we could just inline the deck (decksLibraryContent[id]), but since it's used in two places, we can use `deck` here as well. Otherwise looks good.

KManolov3 (approved):

PR #176, opened 2025-02-19
ci: run all workspaces on changes to global files like package-lock.json

cwillisf:
### Proposed Changes

Treat changes in certain files, such as `/package-lock.json`, as affecting all workspaces.

### Reason for Changes

Without something like this, changes to these files can affect all packages without triggering tests. For example, commit 98932bfc749324e0ae86893103f29ed89dc85f1e broke VM tests, but the problem wasn't flagged by CI because the commit affected only `package-lock.json`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,53 +35,54 @@ jobs:
       - uses: dorny/paths-filter@v3
         id: filter
         with:
+          # Files in the `global` filter affect all workspaces, even if workspace-specific files have not changed.
           filters: |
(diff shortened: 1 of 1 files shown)


KManolov3 (approved):

MiroslavDionisiev (approved):

cwillisf:
VM tests failed as expected (need to undo something that 98932bfc749324e0ae86893103f29ed89dc85f1e did), but the changes to `ci.yml` worked as expected.

PR #196, opened 2025-02-21
Make the account menu more configurable (configurabler?)

georgyangelov:
Previously the menu determined its own state based on the redux session.
Since we don't use that in NGP we want to be able to configure the menu
in a "dumber" way - provide the necessary parameters so that the editor
only gets the values it needs to show/use and doesn't try to determine it
itself.

Part of the story is that the old session data model is different from
the new one. Another part is that the `username` is no longer a unique
identifier (as in a classroom context it may be the actual person name).
And a third part of the story is that some functionality is not (yet)
applicable to NGP, for example "My stuff", logging in from the editor,
account settings, etc.

So we introduce the AccountMenuOptions interface which can be passed
as a prop from the GUI container. If that prop is present - the editor
will use it to determine what to show in the account menu.
If it's not present - then nothing changes - logic is as it was before,
so that we're backwards-compatible with scratch-[www](http://www/).

We also propagate `username` (previously only determined based on session state) & `userOwnsProject` (since usernames are not necessarily unique. Both keep their previous behavior if not provided on the `<GUI` container.


packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -126,6 +128,8 @@ const GUIComponent = props => {
         telemetryModalVisible,
         theme,
         tipsLibraryVisible,
+        username,
+        userOwnsProject,
```
MiroslavDionisiev, line 132: `userOwnsProject` is missing in propTypes

packages/scratch-gui/src/lib/account-menu-options.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    avatarUrl?: string;
+    myStuffUrl?: string;
+    profileUrl?: string;
+    myClassesUrl?: string;
+    myClassUrl?: string;
+    accountSettingsUrl?: string;
+}
+
+export const AccountMenuOptionsPropTypes = PropTypes.shape({
```
MiroslavDionisiev, line 18: `canLogout` and `avatarUrl` are missing in the proptypes

packages/scratch-gui/src/components/menu-bar/menu-bar.jsx
```diff
@@ -968,21 +999,38 @@ const mapStateToProps = (state, ownProps) => {
         loginMenuOpen: loginMenuOpen(state),
         modeMenuOpen: modeMenuOpen(state),
         projectTitle: state.scratchGui.projectTitle,
-        sessionExists: state.session && typeof state.session.session !== 'undefined',
         settingsMenuOpen: settingsMenuOpen(state),
-        username: user ? user.username : null,
-        userOwnsProject: ownProps.authorUsername && user &&
-            (ownProps.authorUsername === user.username),
+        username: ownProps.username ?? (user ? user.username : null),
+        userIsEducator: state.session && state.session.permissions && state.session.permissions.educator,
```
MiroslavDionisiev: Can't we use the permission variable from above

georgyangelov:
cc @cwillisf for context

MiroslavDionisiev (approved):

packages/scratch-gui/src/components/gui/gui.jsx
georgyangelov: Ah, thanks

packages/scratch-gui/src/lib/account-menu-options.ts
georgyangelov: Nice catch

georgyangelov:
@cwillisf Miro has tested this with scratch-www. If you're ok with the change we plan on merging this on Monday

packages/scratch-gui/src/lib/cloud-manager-hoc.jsx
```diff
@@ -111,12 +111,16 @@ const cloudManagerHOC = function (WrappedComponent) {
...
                 projectId,
-                username,
                 hasCloudPermission,
                 isShowingWithId,
                 onShowCloudInfo,
                 /* eslint-enable no-unused-vars */
+
                 vm,
+
+                // Intentionally propagating this one since it's used in the
```
cwillisf: ...since it's used in the... _aaaaarrgh_

> Perhaps he was dictating?

packages/scratch-gui/src/lib/account-menu-options.ts
```diff
@@ -0,0 +1,31 @@
...
+    canHaveSession: PropTypes.bool.isRequired,
```
cwillisf, line 19: In what scenarios would `canHaveSession` be false? Is that for, say, Scratch Lab?

cwillisf (approved):
I saw one comment that seems to accidentally

But the code looks good to me, and your explanation makes sense :+1:

packages/scratch-gui/src/lib/account-menu-options.ts
georgyangelov: I believe so, yes. If this is false then neither "Sign up / sign in", nor "username / log out" is shown

packages/scratch-gui/src/lib/cloud-manager-hoc.jsx
georgyangelov: whoops 😄

georgyangelov: Took me a while but I eventually saw what you

in your other comment 😄

packages/scratch-gui/src/lib/account-menu-options.ts
georgyangelov: This was originally called "sessionExists" but it seems like the session in www always exists, even if you're not logged in, so I renamed it here

PR #212, opened 2025-03-03
Build script improvements

cwillisf:
### Proposed Changes

Changes to some of the build scripts:

- Add a top-level `clean` script and audit `clean` scripts for each workspace (some were missing, some were incomplete)
- Explicitly list packages under `workspaces` to control the order of (for example) `npm --workspaces run ...`
- Now that workspaces are ordered, simplify `npm run build`
- Rebuild the version update

### Reason for Changes

Most of this is just about developer ergonomics, but the version update also improves correctness. This approach ensures that several things happen:

| update for | in scope-specific `package.json` | in `package-lock.json` (global) |
| --- | --- | --- |
| root workspace (optional) | added with new script | added with new script |
| each package | already worked correctly | added with new script |
| package direct dependencies | already worked correctly | added with new script |
| package indirect dependencies | N/A | added with new script |

Note that "package indirect dependencies" includes updating the version of indirect internal dependencies and updating indirect external dependencies if necessary. That can mean relatively complex `package-lock.json` changes that are difficult to predict or simulate with something like `jq`.

The details of the version update process are a little goofy and documented in the script itself.

### Test Coverage

Tested locally with `npm run clean`, `npm run build`, and various `npm version` commands. Since the `npm run build` command is (now) used by CI, it's also tested by posting this PR.


package.json
```diff
@@ -11,21 +11,27 @@
   },
   "scripts": {
     "refresh-gh-workflow": "ts-node scripts/build-gha-workflows.ts",
-    "build-monorepo": "bash ./scripts/build-monorepo.sh",
+    "build-monorepo": "cross-env-shell ./scripts/build-monorepo.sh",
```
cwillisf, line 14: I just learned about `cross-env-shell` this week :sweat_smile:

scripts/npm-version.sh
```diff
@@ -0,0 +1,109 @@
...
+for workspace in "." "${workspace_locations[@]}"; do
+    for dependency in "${workspace_names[@]}"; do
+        update_dependency_in_workspace "$workspace" "$dependency"
+    done
+done
+
+echo "${me}: Asking npm to clean up the lock file..." >&2
+npm install --offline --no-audit --no-fund --ignore-scripts --package-lock-only
+# Sometimes it makes further changes the second time
+npm install --offline --no-audit --no-fund --ignore-scripts --package-lock-only
```
MiroslavDionisiev, line 99: we don't ask how npm works

MiroslavDionisiev (approved):

PR #221, opened 2025-03-24
Deploy to npm tag from branch

cwillisf:
### Proposed Changes

Main change: allow the "Publish" workflow to publish to a tag other than `latest`, depending on the branch name and the state of the "Prerelease" checkbox in GitHub's release UI.

Additional changes:

- `commitlint` now runs as a separate workflow instead of as part of the CI workflow, so you can tell the difference between a `commitlint` failure vs. a build/test failure.
- the release commit message now follows Conventional Commit style.

### Reason for Changes

This allows the `spork` branch to publish to the `spork` tag on npm, making it easier to test on staging. Other branches are likely to want this soon, so I pulled this out of the `spork` branch so we don't need to wait for all the `spork` QA to finish before those other branches can use this change.

### Test Coverage

I've already used this workflow from the `spork` branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,6 @@ jobs:
         with:
           cache: 'npm'
           node-version-file: '.nvmrc'
-      - uses: wagoid/commitlint-github-action@9763196e10f27aef304c9b8b660d31d97fce0f99 # v5
       - name: Debug info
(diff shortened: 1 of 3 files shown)


cwillisf:
Closing in favor of scratchfoundation/scratch-editor#222

PR #228, opened 2025-04-09
[IPR-350] Added no-global-reset to modal component

Bogomil-Stoyanov:
**Ticket:**
[IPR-350](https://scratchfoundation.atlassian.net/jira/software/projects/IPR/boards/90?selectedIssue=IPR-350)

Changes:
А small change to the `ModalComponent` in the `modal.jsx` file - adds a new class `'no-global-reset'` to the modal's class names to ensure the modal is not affected by any global CSS resets in relation to Tailwind CSS reset for scratch-platform.

[IPR-350]: https://scratchfoundation.atlassian.net/browse/IPR-350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/scratch-gui/src/components/modal/modal.jsx
+++ b/packages/scratch-gui/src/components/modal/modal.jsx
@@ -17,7 +17,8 @@ const ModalComponent = props => (
     <ReactModal
         isOpen
         className={classNames(styles.modalContent, props.className, {
-            [styles.fullScreen]: props.fullScreen
+            [styles.fullScreen]: props.fullScreen,
+            'no-global-reset': true
         })}
         contentLabel={props.contentLabel}
         overlayClassName={styles.modalOverlay}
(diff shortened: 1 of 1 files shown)


georgyangelov:
@cwillisf @KManolov3 @MiroslavDionisiev We want to add this class name so that we can reset the CSS reset that NGP uses. Here is the NGP counterpart PR: https://github.com/scratchfoundation/scratch-platform/pull/286 . It's specifically added to modals because currently we can't distinguish editor modals from other (NGP) modals (as they are rendered outside of the editor container).

yo-dawg-reset

The class name is up for debate, I'm wondering if it should be something more generic like "editor-modal". Any input?

cwillisf (approved):
Careful not to cause an ouroboreset!

This seems reasonable... I don't have a strong opinion on the class name until we use it in at least two places. If/when we do that, maybe the commonality between the two will indicate what the name should be. Of course, I also don't mind if you want to change it now.

I do think it would help to add a comment explaining generally why the class is here, especially considering the meaning is in another ~castle~ repo.

YoanBabulkov (approved):

PR #229, opened 2025-04-10
Track asset provenance when using the paint editor

georgyangelov:
### Resolves

Something we talked about on the NGP planning

### Proposed Changes

_Describe what this Pull Request does_

Tracks which asset id an asset edited through the paint editor is based on. Tries to only track uploaded (clean) ones. That's what clean means, right?

Requires the change from https://github.com/scratchfoundation/scratch-storage/pull/754

Note on the unused `store` parameter - in NGP we're not using the webHelper for storing (as it doesn't support async functions), so we can just utilize that parameter there to send to the server together with the rest of the asset.

### Reason for Changes

@cwillisf was talking about it and I just opened the code to see how hard it would be, and there you have it - this PR 😄 I haven't tested it and there is no test coverage, so it's left in a Draft state.

### Test Coverage

I didn't 😅

### DIFF
--- a/packages/scratch-gui/src/containers/paint-editor-wrapper.jsx
+++ b/packages/scratch-gui/src/containers/paint-editor-wrapper.jsx
@@ -24,19 +24,28 @@ class PaintEditorWrapper extends React.Component {
         this.props.vm.renameCostume(this.props.selectedCostumeIndex, name);
     }
     handleUpdateImage (isVector, image, rotationCenterX, rotationCenterY) {
+        const currentCostume = this.props.vm.editingTarget.getCostumes()[this.props.selectedCostumeIndex];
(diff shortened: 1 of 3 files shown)


cwillisf (commented):
Wow, nice!
If I make many edits to a costume before a single project save, would this end up setting that the `provenance` field each time? I think it should only include IDs that made it to the asset server (or to disk, in the apps case)

georgyangelov:
> I think it should only include IDs that made it to the asset server (or to disk, in the apps case)

Yep, that's what I was trying to do with...

```const provenance = currentAsset.clean ? currentAsset.assetId : currentAsset.provenance;```

...so that the provenance is only taken from clean (saved) assets and propagated through unsaved versions.

But as it is untested, there may be some case I haven't considered.

PR #240, opened 2025-04-24
feat: [UEPR-209] release standalone gui as separate package

MiroslavDionisiev:
### Resolves

[UEPR-209](https://scratchfoundation.atlassian.net/browse/UEPR-209)

- added new step in publish workflow for releasing standalone gui as a separate package
- in scratch-gui's scripts added commands for building gui with a build, dist and dist-standalone outputs independently

[UEPR-209]: https://scratchfoundation.atlassian.net/browse/UEPR-209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


georgyangelov (approved):
Nice!

.github/workflows/publish.yml
```diff
@@ -72,26 +72,46 @@ jobs:
...
+        run: |
+          npm run build --workspace @scratch/scratch-vm
+          npm publish --access=public --tag="${{steps.npm_tag.outputs.npm_tag}}" --workspace=@scratch/scratch-vm
         env:
           NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
 
       - name: Publish scratch-gui
-        run: npm publish --access=public --tag="${{steps.npm_tag.outputs.npm_tag}}" --workspace=@scratch/scratch-gui
+        run: |
+          npm run build:dist --workspace @scratch/scratch-gui
```
KManolov3, line 106: We could also remove the `./standalone` export here, as with these changes it will be present as an export, but it won't be built + packaged in the dist.

.github/workflows/publish.yml
```diff
@@ -72,26 +72,46 @@ jobs:
...
+      - name: Publish scratch-gui-standalone
+        run: |
+          jq '
+            .name = "@scratch/scratch-gui-standalone" |
+            del(.peerDependencies) |
+            .exports."." = .exports."./standalone" |
+            del(.exports."./standalone")
+          ' ./packages/scratch-gui/package.json | npx sponge ./packages/scratch-gui/package.json
+
+          npm i
```
cwillisf: Is this `npm i` just to copy the modified values into the lockfile, or does it do more?
In other words, would it be appropriate to replace this with either `npm ci` or `npm i --package-lock-only`?

.github/workflows/publish.yml
```diff
@@ -72,26 +72,46 @@ jobs:
...
+          npm run clean && npm run build:dist-standalone --workspace @scratch/scratch-gui-standalone
```
cwillisf: Should the first part of this be `npm --workspace=@scratch/scratch-gui-standalone run clean`?

KManolov3 (approved):

PR #244, opened 2025-04-28
Fix/uepr 224 svg vulnerabilities

KManolov3:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-224

### Proposed Changes

Handle more cases when sanitizing SVGs

### Reason for Changes

There were reports of multiple ways to exploit svgs by including urls into them, which the browser then loads. 

### Test Coverage

Added a test case with multiple ways of adding svg urls


MiroslavDionisiev (approved):

yuri-kiss:
yay, the issue I found is being fixed :D

PR #262, opened 2025-07-08
RaspberryPiFoundation feature enhancements

sebjacobs:
### Proposed Changes

This pull request includes a number of changes from the RaspberryPiFoundation fork of scratch-gui:

* Remove duplicate `remixProject` export
* Allow the MenuBar to be hidden (by providing the `menuBarHidden` prop).
* Export `manualUpdateProject` reducer action.
* Export `setStageSize` reducer action.

### Reason for Changes

We maintain a number of forks of scratch-gui across the RaspberryPiFoundation which we are looking at rationalising. We believe that these particular changes could be useful for other people and therefore would be keen to get these features merged upstream.

### Test Coverage

We were unable to find an existing unit test for the Gui component and so we have not added any test coverage around the new `menuBarHidden` prop.


sebjacobs:

sebjacobs:
recheck

packages/scratch-gui/src/index.ts
```diff
@@ -4,7 +4,6 @@ import {legacyConfig} from './legacy-config';
 export {default} from './containers/gui.jsx';
 export {default as GUIComponent} from './components/gui/gui.jsx';
 export {default as AppStateHOC} from './lib/app-state-hoc.jsx';
-export {remixProject} from './reducers/project-state.js';
```
cwillisf, line 7: Was this removal intentional? If so, could you explain why?

cwillisf (commented):
One question about `remixProject`; if there's a good reason for that change, then this looks great! Otherwise, it looks 99% great ;)

Either way, thank you for sharing your work!

packages/scratch-gui/src/index.ts
sebjacobs: @cwillisf this action is being exported twice, once on this line (line 7) and also on line 13 via `exported-reducers.ts`:
* https://github.com/scratchfoundation/scratch-editor/blob/develop/packages/scratch-gui/src/index.ts#L13
* https://github.com/scratchfoundation/scratch-editor/blob/develop/packages/scratch-gui/src/exported-reducers.ts#L33

sebjacobs:
> One question about `remixProject`; if there's a good reason for that change, then this looks great! Otherwise, it looks 99% great ;)
> 
> Either way, thank you for sharing your work!

@cwillisf thanks for taking a look. Sorry for the confusion I explained my reasoning in the individual commits but did not mention them in the PR description. I have also responded to your query inline on the diff.

packages/scratch-gui/src/index.ts
cwillisf: Gotcha - thanks! (I should have looked at the commit -- sorry about that :sweat_smile:)

cwillisf (approved):

cwillisf:
Sorry for the delay, and thanks again!

PR #264, opened 2025-07-15
[feature] UEPR-252 Implement manual update of project thumbnails

KManolov3:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-252

### Proposed Changes

Update logic, so that thumbnails are now updated manually

### Reason for Changes

This will hopefully allow us to reduce monthly costs connected with thumbnail updates


packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -128,6 +129,7 @@ const GUIComponent = props => {
         onTelemetryModalCancel,
         onTelemetryModalOptIn,
         onTelemetryModalOptOut,
+        onUpdateProjectThumbnail,
```
georgyangelov, line 132: Thumbnail saves are currently done through the Storage interface - there is a `saveProjectThumbnail` method that is currently called. Is that still called to save the thumbnail to the server? And this is just to notify that the user saved it, right?

packages/scratch-gui/src/lib/store-project-thumbnail.js
```diff
@@ -0,0 +1,20 @@
+import log from './log';
+
+export const storeProjectThumbnail = (vm, callback) => {
+    try {
+        getProjectThumbnail(vm, callback);
```
georgyangelov, line 5: "store" calls "get"? Hm?

Does this also save the project thumbnail to the backend?

packages/scratch-gui/src/components/gui/gui.jsx
KManolov3: As discussed offline, the store's `saveProjectThumbnail` is used as fallback when the onUpdate callback isn't passed - which is the case in NGP

packages/scratch-gui/src/lib/store-project-thumbnail.js
KManolov3: No, the callback saves it, the vm makes a snapshot of the current state.

KManolov3:
TODO: revert c48f364ed08aae873469ff0dcc7924574e61b9df, as passing the Storage.saveThumbnail default in the projectSaverHOC should be enough to propagate it to the stage header

MiroslavDionisiev (approved):

MiroslavDionisiev (approved):

PR #267, opened 2025-07-18
Improve GHA workflows

cwillisf:
### Proposed Changes

Various improvements to the GHA workflows, especially for the build-and-test workflow.

- Consider `.nvmrc` when caching `node_modules`, so we don't (for example) load Node 20 modules when we switch to Node 22
- Move `commitlint` to its own separate workflow, so it doesn't block the build-and-test workflow
- Move tests into a separate job matrix so it's easier to tell what fails and re-run them if necessary
- Move path filters out of `ci.yml` (where the YAML must be embedded in a string) and into a separate YAML file (where they can be actual YAML and your editor can provide proper syntax highlighting)

At some point in the future, it'd be nice to matrix out the builds as well, but we'll need to eliminate cases where the build output of one package is an input to another.

### Test Coverage

Runs existing tests

### DIFF
--- a/.github/actions/install-dependencies/action.yml
+++ b/.github/actions/install-dependencies/action.yml
@@ -33,7 +33,7 @@ runs:
           packages/*/node_modules
           packages/*/src/generated
           packages/scratch-gui/static/microbit
-        key: ${{ runner.os }}-npm-${{ hashFiles('package-lock.json') }}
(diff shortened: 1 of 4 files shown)


MiroslavDionisiev (approved):

KManolov3 (approved):

KManolov3:
This looks great!

PR #290, opened 2025-08-11
[IPR-979] Omit multiple unnecessary project title update calls

sashodada:
### Proposed Changes

- only handle default project title update once, and do not invoke `onUpdateProjectTitle` to avoid title updates for nonexistent project
- only call `onUpdateProjectTitle` if the title has changed AND the project has loaded

### Reason for Changes

There are multiple calls to `onUpdateProjectTitle` while a project isn't actually loaded (e.g. the editor is in some "creating new" state). This causes multiple unnecessary and currently failing requests in WWW and side effects in other projects that integrate the editor.


KManolov3:
Note to self: test this with www before next release

KManolov3 (approved):

georgyangelov (approved):
Looks good from my perspective, but we should ping CWF just in case.

PR #291, opened 2025-08-12
[UEPR-290] Update face sensing configuration

adzhindzhi:
### Resolves

[UEPR-290](https://scratchfoundation.atlassian.net/browse/UEPR-290)

### Proposed Changes

- Install @mediapipe/face_detection package through npm
- Copy mediapipe files into the output folder during build
- Add fallback logic to load mediapipe from CDN if local files are missing

### Reason for Changes

Currently the path for `@mediapipe/face_detection` uses a CDN link - loading the package without needing to install it locally. If  the library version changes with a breaking change it will stop working without any visible reason.

### Test Coverage

No test required

[UEPR-290]: https://scratchfoundation.atlassian.net/browse/UEPR-290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/scratch-vm/src/extensions/scratch3_face_sensing/index.js
+++ b/packages/scratch-vm/src/extensions/scratch3_face_sensing/index.js
@@ -41,16 +41,26 @@ class Scratch3FaceSensingBlocks {
         const model = FaceDetection.SupportedModels.MediaPipeFaceDetector;
         const detectorConfig = {
             runtime: 'mediapipe',
-            solutionPath: 'https://cdn.jsdelivr.net/npm/@mediapipe/face_detection',
+            solutionPath: '/chunks/mediapipe/face_detection',
             maxFaces: 1
         };
     
-        FaceDetection.createDetector(model, detectorConfig).then(detector => {
-            this.faceDetector = detector;
-            if (this.runtime.ioDevices) {
-                this._loop();
-            }
-        });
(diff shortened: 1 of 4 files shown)


packages/scratch-vm/src/extensions/scratch3_face_sensing/index.js
```diff
@@ -41,16 +42,26 @@ class Scratch3FaceSensingBlocks {
...
     
-        FaceDetection.createDetector(model, detectorConfig).then(detector => {
-            this.faceDetector = detector;
-            if (this.runtime.ioDevices) {
-                this._loop();
-            }
-        });
+        FaceDetection.createDetector(model, detectorConfig)
+            .catch(() => {
+                const fallbackConfig = {
```
KManolov3, line 51: nice!

KManolov3 (approved):
LGTM. If both the local and remote configs fail, we'll "fail silently", but tough luck in this case. More generally, it seems like a good idea to expand the webpack config file with comments (what does what, and why we need it) - both for maintainers of this repo, and, such as in this case, as a guide for other repos which wish to use this one as a library and need to make the necessary adjustments (copying static files, fetch workers, etc..).

MiroslavDionisiev (approved):

Issue #269, opened 2025-07-20
Vector graphics with shadows no longer display


scratching-claws:
### Expected Behavior

imported vector graphics with a shadow, previously displayed correctly, with the shadow, in both the editor and the scratch stage (but not in costume thumbnails)

Any edits would remove the shadow (and similar effects such as corner rounding)

### Actual Behavior

imported vector graphics with a shadow, will not display on the stage (but will in the editor).  As far as I can tell, this is only affected by the shadow property.  Corner rounding still follows the previous behaviour.

### Steps to Reproduce

1. Go to 'https://scratch.mit.edu/create'
2. Click on 'import costume' or 'upload sprite'
3. Import an SVG costume with a shadow effect (try creating one in a tool like Figma)
4. Any element with the shadow effect does not render

See the project: https://scratch.mit.edu/projects/655176140/editor

look at the costumes of the first sprite

### System Details
Cross platform

**Screenshots**

<img width="488" height="478" alt="Image" src="https://github.com/user-attachments/assets/76a46fc9-437c-47ce-96de-8e040a0fc90f" />

<img width="1919" height="967" alt="Image" src="https://github.com/user-attachments/assets/a898ca8c-13c8-47e5-861e-e5cf94c4ec84" />

https://scratch.mit.edu/projects/1198678674

ampelectrecuted:
pretty clear you're using scratchaddons, try disabling all browser extensions and see if it will work

medlans:
> pretty clear you're using scratchaddons, try disabling all browser extensions and see if it will work

I can test it myself with the Rays backdrop (which seemed to have the problem)
Edit
My browser / operating system: Windows 7, Chrome 146.0.0.0, No Flash versions detected
Can replicate with and without extensions with rays

PR #300, opened 2025-08-24
fix: allow SVG filter elements

scratching-claws:
Adds a rule to allow the SVG filter tag, when sanitizing SVGs, so that effects such as shadows and blur do not lead to SVGs which don't diplay on the project stage

### Resolves

Fixes this issue: https://github.com/scratchfoundation/scratch-editor/issues/269

### Proposed Changes

Adds a rule to allow the SVG filter tag

### Reason for Changes

Stops the SVG sanitizer from removing filter tags from SVGs, so that SVGs with shadows/blur display correctly on the stage

### DIFF
--- a/packages/scratch-svg-renderer/src/sanitize-svg.js
+++ b/packages/scratch-svg-renderer/src/sanitize-svg.js
@@ -130,6 +130,8 @@ sanitizeSvg.sanitizeSvgText = function (rawSvgText) {
     let sanitizedText = DOMPurify.sanitize(rawSvgText, {
         USE_PROFILES: {svg: true},
         FORBID_TAGS: ['a', 'audio', 'canvas', 'video'],
+        // Allow SVG filter elements (used for shadows, blur)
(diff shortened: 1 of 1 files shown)


nimeratus:
Will this also allow things like `<feGaussianBlur>` that should be inside the `<filter>` element?

I've looked at DOMPurify's readme page and it says there is a toggle for SVG filters
https://github.com/cure53/DOMPurify?tab=readme-ov-file#control-our-allow-lists-and-block-lists

scratching-claws:
> Will this also allow things like `<feGaussianBlur>` that should be inside the `<filter>` element?
> 
> I've looked at DOMPurify's readme page and it says there is a toggle for SVG filters https://github.com/cure53/DOMPurify?tab=readme-ov-file#control-our-allow-lists-and-block-lists

That is a very good point, and thank you for raising it.  I have changed the pull request so that is now uses the SVG filters profile, instead of only allowing the additional <filter> tag

PR #303, opened 2025-09-01
[UEPR-312] Clean up face sensing codebase

adzhindzhi:
### Resolves

[UEPR-312](https://scratchfoundation.atlassian.net/browse/UEPR-312)

### Proposed Changes

- Delete commented out and unused code
- Remove unused properties from FACE_SENSING_STATE
- Mark private methods and properties with underscore for consistency
- Add method/propery docs
- Extract some generic methods to utils.js
- Other minor improvements

### Reason for Changes

Get the face sensing code ready for production

[UEPR-312]: https://scratchfoundation.atlassian.net/browse/UEPR-312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/scratch-render/src/RenderWebGL.js
+++ b/packages/scratch-render/src/RenderWebGL.js
@@ -1069,6 +1069,16 @@ class RenderWebGL extends EventEmitter {
         return false;
     }
 
+    /**
+     * Determine if the drawable is touching a rectangle.
+     *
+     * @param {int} drawableID The ID of the drawable to check.
+     * @param {int} left - The left X coordinate of the rectangle.
+     * @param {int} top - The top Y coordinate of the rectangle.
+     * @param {int} right - The right X coordinate of the rectangle.
+     * @param {int} bottom - The bottom Y coordinate of the rectangle.
+     * @returns {boolean} If the drawable has any pixels that would draw in the rectangle area
+     */
     drawableTouchingScratchRect (drawableID, left, top, right, bottom) {
         const drawable = this._allDrawables[drawableID];
         if (!drawable) {
@@ -1090,6 +1100,14 @@ class RenderWebGL extends EventEmitter {
         return false;
     }
 
+    /**
+     * Determine if the drawable is touching a point in the Scratch coordinate system
+     *
+     * @param {int} drawableID The ID of the drawable to check.
+     * @param {int} x The x coordinate of the point.
+     * @param {int} y The y coordinate of the point.
+     * @returns {boolean} If the drawable is touching the point
+     */
     drawableTouchingScratchPoint (drawableID, x, y) {
         const drawable = this._allDrawables[drawableID];
         if (!drawable) {
--- a/packages/scratch-vm/src/extensions/scratch3_face_sensing/index.js
+++ b/packages/scratch-vm/src/extensions/scratch3_face_sensing/index.js
@@ -5,7 +5,7 @@ const MathUtil = require('../../util/math-util');
 const formatMessage = require('format-message');
 const Video = require('../../io/video');
 const TargetType = require('../../extension-support/target-type');
-// const Posenet = require('@tensorflow-models/posenet');
+const {distance, toScratchCoords} = require('./utils');
 
 const FaceDetection = require('@tensorflow-models/face-detection');
 const mediapipePackage = require('@mediapipe/face_detection/package.json');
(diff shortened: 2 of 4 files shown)


packages/scratch-vm/src/extensions/scratch3_face_sensing/index.js
```diff
@@ -118,63 +152,62 @@ class Scratch3FaceSensingBlocks {
...
-    }
-
-    static get STATE_KEY () {
-        return 'Scratch.faceSensing';
-    }
 
-    static get DEFAULT_FACE_SENSING_STATE () {
-        return {
-            attachedToPartNumber: null,
-            prevX: 0,
```
KManolov3, line 169: Was the rest of this state (beside `attachedToPartNumber`) end up being unused?

KManolov3 (approved):
Looks much cleaner now :)

packages/scratch-vm/src/extensions/scratch3_face_sensing/index.js
adzhindzhi: Yes, everything else was only used in methods that were only used by the commented out blocks (e.g. `attachToPart`, `updateAttachments`, etc.)

packages/scratch-render/src/RenderWebGL.js
```diff
@@ -1069,6 +1069,16 @@ class RenderWebGL extends EventEmitter {
...
+    /**
+     * Determine if the drawable is touching a rectangle.
+     *
+     * @param {int} drawableID The ID of the drawable to check.
+     * @param {int} left - The left X coordinate of the rectangle.
+     * @param {int} top - The top Y coordinate of the rectangle.
+     * @param {int} right - The right X coordinate of the rectangle.
+     * @param {int} bottom - The bottom Y coordinate of the rectangle.
+     * @returns {boolean} If the drawable has any pixels that would draw in the rectangle area
+     */
```
cwillisf, line 1081: 😎
If you run across missing or incorrect JSDoc that's making your life difficult, please feel very free to fix it up. Some of the JSDoc in here, especially in the VM and renderer, was written in the early days of JSDoc and (more importantly) the early days of our (at least my) understanding of JSDoc and JavaScript in general, so some of it is wrong or invalid, and a lot of it is just missing. 

TL;DR: Thank you!

cwillisf: (I'd like to move to TS when and where it makes sense, but that's a long walk...)

packages/scratch-vm/src/extensions/scratch3_face_sensing/index.js
```diff
@@ -232,7 +264,6 @@ class Scratch3FaceSensingBlocks {
                     blockType: BlockType.COMMAND,
                     filter: [TargetType.SPRITE]
                 },
-                '---',
```
cwillisf: Were these separators intentionally removed? It seems OK to me; just double-checking.

packages/scratch-vm/src/extensions/scratch3_face_sensing/utils.js
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Convert a point to Scratch coordinates.
+ *
+ * @param {{x: number, y: number}} position - Original coordinates of the point.
+ * @returns {{x: number, y: number}} Converted point in Scratch coordinates.
+ */
+const toScratchCoords = function (position) {
+    return {
+        x: position.x - 240,
+        y: 180 - position.y
```
cwillisf, line 24: 🙃

packages/scratch-vm/src/sprites/rendered-target.js
```diff
@@ -756,7 +756,7 @@ class RenderedTarget extends Target {
      * Return whether touching a point.
      * @param {number} x X coordinate of test point.
      * @param {number} y Y coordinate of test point.
-     * @return {boolean} True iff the rendered target is touching the point.
+     * @return {boolean} True if the rendered target is touching the point.
```
cwillisf: I'm OK with these changes, but just as an FYI, I think in each of these cases the "iff" was intended as an abbreviation for ["if and only if"](https://en.wikipedia.org/wiki/Iff) rather than being a typo. On the other hand, that's excessively fancy and maybe not even strictly true. Also—and this may be the strongest argument in favor of this change—red squiggly lines are annoying. :D

cwillisf (approved):

packages/scratch-vm/src/extensions/scratch3_face_sensing/index.js
ericrosenbaum: The separators were there by design and should not be removed - please put them back

ericrosenbaum:
While we're fixing comments - in the extension index.js line 28 `Class for the motion-related blocks in Scratch 3.0` should be fixed to say "Face Sensing blocks"

ericrosenbaum:
This is great so far - there are actually a few additional bits of code that can be removed. They are all related to the idea of "attachment," which I prototyped but decided not to use. 

If you're interested, I talk about the different ideas we tried, including an "attach" block, and how we settled on the final design, in [this talk](https://youtu.be/d7Uzd3VFzg4?t=360) from last year's Blockly summit. 

The stop button event is wired up to `_clearAttachments` and that function uses `_getFaceSensingState` which uses `DEFAULT_FACE_SENSING_STATE`. All of that can be removed.

ericrosenbaum (changes requested):
I've suggested a couple of changes - they could also be done in a follow-up PR.

ericrosenbaum (changes requested):
The one last thing I found to remove is the `STATE_KEY`. Thanks!

PR #307, opened 2025-09-06
Fix dragging monitors on small stage

nimeratus:
### Resolves

_What Github issue does this resolve (please include link)?_

* https://github.com/scratchfoundation/scratch-gui/issues/2949
* https://github.com/scratchfoundation/scratch-gui/issues/2367
* https://github.com/scratchfoundation/scratch-gui/issues/9756
* https://github.com/scratchfoundation/scratch-gui/issues/9610

(most of these are duplicates of each other)

### Proposed Changes

_Describe what this Pull Request does_

1. Resize the dragging bounds with the same CSS transform that resizes the monitors, instead of changing the width/height CSS properties
2. Remove the `<Box>` that used to do the CSS transform
3. Set the `scale` property on the `<Draggable>`
4. Prevent the monitor from going to fractional coordinates even if `scale` is a fraction

### Reason for Changes

1. This way CSS word wrapping and react-draggable will let the monitors go to the edges of the stage
2. It doesn't do the transform anymore and I don't think it did anything else
3. This way react-draggable knows the speed at which it should move the monitors
4. The existing code expects the monitor coordinates to be integers [here](https://github.com/scratchfoundation/scratch-editor/blob/440957a0c7b4f76ebef93c2dc8706a5e413bd7e6/packages/scratch-gui/src/containers/monitor.jsx#L107-L108) and there might be other places where it expects integers

### Test Coverage

I've run `npm test` and it accepted it (I think) (it didn't say "error" like when it didn't like the identation)

I've tried it out in these browsers and it seems to work

* Windows 10, Chrome 139
* Android, Chrome 138

### DIFF
--- a/packages/scratch-gui/src/components/monitor-list/monitor-list.jsx
+++ b/packages/scratch-gui/src/components/monitor-list/monitor-list.jsx
@@ -22,6 +22,7 @@ const MonitorList = props => (
                 .map(monitorData => (
                     <Monitor
                         draggable={props.draggable}
+                        scale={props.stageSize.scale}
                         height={monitorData.height}
                         id={monitorData.id}
                         isDiscrete={monitorData.isDiscrete}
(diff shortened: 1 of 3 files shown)


nimeratus:
me:

> (by the way if you wonder why all of my commit titles starts with "fix:", I think there is an automatic test somewhere that checks if commit titles start with something like that and I don't want my PR to fail just because of the commit titles)

commitlint:

> ✖   body's lines must not be longer than 100 characters [body-max-line-length]

Noooo, why...

PR #314, opened 2025-09-11
Fix dragging monitors on small stage

nimeratus:
### Resolves

* https://github.com/scratchfoundation/scratch-gui/issues/2949
* https://github.com/scratchfoundation/scratch-gui/issues/2367
* https://github.com/scratchfoundation/scratch-gui/issues/9756
* https://github.com/scratchfoundation/scratch-gui/issues/9610
(most of these are duplicates of each other)

There is [a PR that solves this issue](https://github.com/scratchfoundation/scratch-gui/pull/5795) in scratch-gui but I think this way it can be solved with much less changes (out of the 42 added rows 22 is just changing white space for eslint, only the other 20 rows are actual changes)

### Proposed Changes

1. Resize the dragging bounds with the same CSS transform that resizes the monitors, instead of changing the width/height CSS properties
    - This way CSS word wrapping and react-draggable will let the monitors go to the edges of the stage
2. Remove the `<Box>` that used to do the CSS transform
    - It doesn't do the transform anymore and I don't think it did anything else
3. Set the `scale` property on the `<Draggable>`
    - This way react-draggable knows the speed at which it should move the monitors
4. Prevent the monitor from going to fractional coordinates even if `scale` is a fraction
    - The existing code expects the monitor coordinates to be integers [here](https://github.com/scratchfoundation/scratch-editor/blob/440957a0c7b4f76ebef93c2dc8706a5e413bd7e6/packages/scratch-gui/src/containers/monitor.jsx#L107-L108) and there might be other places where it expects integers
    1. set the `grid` property on the `<Draggable>`
        - round coordinates in CSS transform so the monitor appears to move to the position where Scratch will think it is
    2. round the coordinates in the `handleDragEnd` event handler
        - the `<Draggable>` doesn't round the coordinates for the event handler

### Reason for Changes

explained next to each change in Proposed Changes

### Test Coverage

I've run `npm test` and it accepted it (I think) (it didn't say "error" like when it didn't like the identation)

I've tried it out in these browsers and it seems to work

* Windows 10, Chrome 139
* Android, Chrome 138

I've temporarily added this to `handleDragEnd` for testing and it didn't cause any popups:
```js
if(this.element.style.transform!==`translate(${Math.round(x)}px, ${Math.round(y)}px)`) {
    alert(`monitor position doesn't match CSS: [${x},${y}] - ${this.element.style.transform}`);
}
```

### DIFF
--- a/packages/scratch-gui/src/components/monitor/monitor.jsx
+++ b/packages/scratch-gui/src/components/monitor/monitor.jsx
@@ -52,6 +52,7 @@ const MonitorComponent = props => (
             defaultClassNameDragging={styles.dragging}
             disabled={!props.draggable}
             onStop={props.onDragEnd}
+            scale={props.scale} // Used for moving the monitor at the correct speed
         >
             <Box
                 className={styles.monitorContainer}
@@ -138,6 +139,7 @@ MonitorComponent.propTypes = {
     category: PropTypes.oneOf(Object.keys(categoryColorMap)),
     componentRef: PropTypes.func.isRequired,
     draggable: PropTypes.bool.isRequired,
+    scale: PropTypes.number,
     label: PropTypes.string.isRequired,
     mode: PropTypes.oneOf(monitorModes),
     onDragEnd: PropTypes.func.isRequired,
--- a/packages/scratch-gui/src/components/monitor/monitor.jsx
+++ b/packages/scratch-gui/src/components/monitor/monitor.jsx
@@ -53,6 +53,8 @@ const MonitorComponent = props => (
             disabled={!props.draggable}
             onStop={props.onDragEnd}
             scale={props.scale} // Used for moving the monitor at the correct speed
+            grid={[1 * props.scale, 1 * props.scale]} // Tell the Draggable to round the coordinates
+            // See https://github.com/react-grid-layout/react-draggable/issues/664 on why it's not just [1, 1]
         >
             <Box
                 className={styles.monitorContainer}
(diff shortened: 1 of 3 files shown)


packages/scratch-gui/src/components/monitor/monitor.jsx
nimeratus, line 44: This has been updated since I made the pull request, but GitHub thinks that it was updated 8 months ago (before I made the PR) and it shows the old version here. Does this mean that the PR is broken?

mxmou: The PR is fine, as indicated by the status at the bottom of the page:
image

Typically, PRs can be merged even if there are newer changes on the upstream branch, unless there's a [merge conflict](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/addressing-merge-conflicts/about-merge-conflicts).

nimeratus: But why does it say when I go to [the monitor.jsx file](https://github.com/scratchfoundation/scratch-editor/blob/develop/packages/scratch-gui/src/components/monitor/monitor.jsx) that it was last changed 8 months ago, when it changed less than 1 month ago, because it no longer uses the  `<ContextMenuTrigger>` that it used when I made the PR

mxmou: Your branch is based on the Scratch code as it was 8 months ago. It's possible to merge Scratch's newer changes back into it, but there's no reason to do that if there are no conflicts.

nimeratus: I've made [a test](https://github.com/nimeratus/scratch-editor/pull/7) and it looks like the date says when the code was written and not when it was merged, so that's why it looked so strange.

nimeratus:
The CLA Document has an error: scratchfoundation/scratch-agreements#7

PR #325, opened 2025-09-23
feat: [IPR-1063] Add locale to editor state props

MiroslavDionisiev:
### Resolves
[IPR-1063](https://scratchfoundation.atlassian.net/browse/IPR-1063)

### Proposed Changes
Add locale to the editor state params that is passed from NGP. If such is not passed keep the old old approach of getting locale

[IPR-1063]: https://scratchfoundation.atlassian.net/browse/IPR-1063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/scratch-gui/src/lib/editor-state.tsx
+++ b/packages/scratch-gui/src/lib/editor-state.tsx
@@ -21,6 +21,7 @@ export interface EditorStateParams {
     isPlayerOnly?: boolean;
     showTelemetryModal?: boolean;
     isEmbedded?: boolean;
+    locale?: string;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/scratch-gui/src/lib/editor-state.tsx
```diff
@@ -40,10 +41,18 @@ export class EditorState {
         let enhancer;
 
         let initializedLocales = localesInitialState;
-        const locale = detectLocale(Object.keys(locales));
+
+        let locale = 'en';
+        if (params.locale && Object.keys(locales).includes(params.locale)) {
```
georgyangelov: I remember we talked about having a `console.warn` if `Object.keys(locales).includes(params.locale)` is false.

georgyangelov (approved):

packages/scratch-gui/src/lib/editor-state.tsx
MiroslavDionisiev: I forgot about it. Will add it

KManolov3 (approved):

georgyangelov (approved):

PR #327, opened 2025-09-24
Update to eslint v9 and eslint-config-scratch v12

cwillisf:
### Proposed Changes

- Update several dependencies:
    - `eslint`: v8 to v9
    - `eslint-config-scratch`: v9 to v12
    - `jest` and `ts-jest`: v21 to v25
      - updated test snapshots (new `jest` doesn't record `undefined` properties, etc.)
    - various support / plugin updates as well
- Remove legacy eslint configuration files and create new, flat ones
- Enable TypeScript & type-aware linting in `scratch-gui`
- Fix tests broken by all these updates

### Reason for Changes

- Work around typescript-eslint/typescript-eslint#11455, which is currently causing Renovate lock file updates to fail
- Keeping up to date is generally good

### Test Coverage

Covered by existing tests, which have been updated to accommodate these changes.


packages/scratch-gui/src/lib/cloud-provider.js
```diff
@@ -99,6 +99,11 @@ class CloudProvider {
         return Math.random() * t;
     }
 
+    /**
+     * Set a timeout for a function to be called after a delay.
+     * @param {Function} fn The function to call after the delay.
+     * @param {number} time The delay time in milliseconds.
```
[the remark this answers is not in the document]
cwillisf: I added a jsdoc comment specifically to make TS linting happy about the arg types

packages/scratch-gui/src/lib/legacy-storage.ts
```diff
@@ -69,6 +69,9 @@ export class LegacyStorage implements GUIStorage {
         vmState: string,
         params: { originalId: string; isCopy: boolean; isRemix: boolean; title: string; }
     ): Promise<{ id: string | number; }> {
+        if (!this.projectHost) {
+            return Promise.reject(new Error('Project host not set'));
+        }
```
cwillisf, line 74: Without this, TS said that passing `this.projectHost` to `saveProjectToServer` was invalid because it might be `undefined`.

packages/scratch-gui/test/unit/util/project-fetcher-hoc.test.jsx
```diff
@@ -60,9 +60,10 @@ describe('ProjectFetcherHOC', () => {
         );
         storage.load = originalLoad;
         // nextTick needed since storage.load is async, and onFetchedProject is called in its then()
-        process.nextTick(
+        setTimeout(
```
cwillisf, line 74: For some reason, `process.nextTick` wasn't delaying long enough. Waiting for 1ms seems to work. It wouldn't surprise me if this turns out to be flaky, though. These tests should ideally be event-driven (watch for a "load finished" event or something), but that'd be a larger change than I wanted to integrate into this PR.

packages/scratch-gui/package.json
```diff
@@ -165,7 +163,7 @@
     "selenium-webdriver": "3.6.0",
     "semantic-release": "19.0.5",
     "stream-browserify": "3.0.0",
-    "ts-jest": "21.2.4",
+    "ts-jest": "^25.5.1",
```
cwillisf, line 172: I was surprised to find that I had to update `jest` and `ts-jest` to get this all to work. I'm not sure exactly what the interaction was, but once I updated all the `eslint`-related things, `jest` started to fail with a parse error.

packages/scratch-vm/src/extensions/scratch3_video_sensing/index.js
```diff
@@ -354,15 +354,6 @@ class Scratch3VideoSensingBlocks {
...
 
-    /**
-     * States the video sensing activity can be set to.
-     * @readonly
-     * @enum {string}
-     */
-    static get VideoState () {
-        return VideoState;
-    }
-
```
cwillisf, line 365: This was unused and causes a namespace collision for TS

KManolov3:
@cwillisf It's likely going to not be pretty, but should we try pointing this to https://github.com/scratchfoundation/scratch-editor/tree/UEPR-282-face-sensing, so that we are ready when it ends up being merged into develop in the next week or so?

cwillisf:
> @cwillisf It's likely going to not be pretty, but should we try pointing this to [`UEPR-282-face-sensing`](https://github.com/scratchfoundation/scratch-editor/tree/UEPR-282-face-sensing?rgh-link-date=2025-09-29T13%3A56%3A54Z), so that we are ready when it ends up being merged into develop in the next week or so?

Probably a good idea... I'll see what it looks like.

packages/scratch-gui/package.json
```diff
@@ -200,7 +198,8 @@
     "moduleNameMapper": {
       "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)\\??$": "<rootDir>/test/__mocks__/fileMock.js",
       "\\.(css|less)$": "<rootDir>/test/__mocks__/styleMock.js",
-      "editor-msgs(\\.js)?$": "<rootDir>/test/__mocks__/editor-msgs-mock.js"
+      "editor-msgs(\\.js)?$": "<rootDir>/test/__mocks__/editor-msgs-mock.js",
+      "^@scratch/scratch-svg-renderer$": "<rootDir>/../scratch-svg-renderer/src/"
```
KManolov3: Just curious - why did we end up needing this for svg-renderer, but not for e.g. -vm

KManolov3 (approved):
That's quite a few changes! Let's wait for the face sensing release before merging this in develop, just to be sure.

packages/scratch-gui/package.json
cwillisf: I figured it out, at least partially: `ts-jest@21.2.4` requires `typescript@^2.4.1`, but `@typescript-eslint/eslint-plugin` requires `typescript>=4.8.4 <6.0.0`. Npm "helpfully" resolves this by (among other things) providing a newer `jest-cli` to `scratch-gui` than the `jest@21.2.1` that `scratch-gui/package.json` asks for. Bringing in a newer version of Jest introduces several new requirements that seem to be met by properly updating `jest` and `ts-jest`.

There are still a few holes in my understanding here, like why npm makes available a version of `jest` that doesn't match `package.json` instead of throwing an error, but I'm at least _closer_ to understanding it :sweat_smile:

cwillisf:
I just rebuilt this on top of the React 18 and Face Sensing changes. I also ran `eslint --fix` since these changes caused so many new lint warnings. Looks like Renovate did its thing while I had my back turned, but it should only take a little bit more merging before this is ready to go...

PR #330, opened 2025-09-26
[UEPR-331] Implement face sensing callouts

KManolov3:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-331

### Proposed Changes
Add feature callouts for new face sensing feature

### Reason for Changes

To introduce the face sensing feature to users.


packages/scratch-gui/package.json
cwillisf: Probably already on your todo list: this will need a version bump, probably to `3.1.0`, once scratchfoundation/scratch-webpack-configuration#105 gets merged

cwillisf (changes requested):
Looks good other than waiting for scratchfoundation/scratch-webpack-configuration#105

One idea to consider: it might make sense to make the callouts appear as long as `now` is less than some end date. That way, if we release this in the apps, the callouts will appear for a while and then go away even if the user doesn't update their copy of the app. On the other hand, maybe we want that date to be different on the web vs. in the apps. Hmm...

packages/scratch-gui/src/components/library/library.jsx
```diff
@@ -144,11 +172,57 @@ class LibraryComponent extends React.Component {
...
+        }
+    }
+    componentWillUnmount () {
+        if (this.driver) {
+            this.driver.destroy();
+            this.driver = null;
+        }
     }
     handleSelect (id) {
+        if (this.state.shouldShowFaceSensingCallout && !this.driver) {
```
KManolov3: If a driver doesn't exist we don't want to do anything for the callout, but we'd still want to call the props' `onItemSelected`. The same issue exists in `handleExtensionButtonClick`.

packages/scratch-gui/src/components/library-item/library-item.raw.css
```diff
@@ -0,0 +1,53 @@
...
+.driver-active:has(.tooltip-face-sensing-modal) > .driver-overlay {
+    visibility: hidden;
+}
+
+/* Fallback, if :has is not supported */
+.tooltip-face-sensing-modal ~ .driver-overlay {
+    visibility: hidden;
+}
+
+:not(body):has(>.driver-active-element) {
```
KManolov3: Assuming the user's browser doesn't support :has, what would be the behaviour here?

packages/scratch-gui/src/components/extension-button/extension-button.jsx
```diff
@@ -0,0 +1,169 @@
...
+        intl,
+        showNewFeatureCallouts,
+        onExtensionButtonClick,
+        username
+    } = props;
+
+    const driverRef = useRef(null);
+    // Keep in a state to avoid reads from localStorage on every render.
+    const [shouldShowFaceSensingCallouts, setShouldShowFaceSensingCallouts] =
+        useState(showNewFeatureCallouts && !hasIntroducedFaceSensing(username));
```
KManolov3: If the user has selected the face sensing extension through the extensions modal, and 'hasUsedFaceSensing' is set, we wouldn't want to show the callout here.

packages/scratch-gui/src/components/library-item/library-item.raw.css
adzhindzhi: Nice catch, I'll try to think of a fallback selector. Otherwise, the scroll in the extensions modal is not displayed, which cuts off a few of the extension items.

packages/scratch-gui/src/components/library/library.jsx
```diff
@@ -196,24 +200,40 @@ class LibraryComponent extends React.Component {
...
+            }
+            
+            this.animationFrameId = null;
+        });
     }
     handleSelect (id) {
-        if (this.state.shouldShowFaceSensingCallout && !this.driver) {
+        const selectedItem = this.getFilteredData().find(item => this.constructKey(item) === id);
+        
+        if (this.state.shouldShowFaceSensingCallout && !this.driver && selectedItem.extensionId !== 'faceSensing') {
```
KManolov3, line 234: When would this happen? On first click?

adzhindzhi: Yes, when the first click is on one of the other extensions

packages/scratch-gui/src/components/library-item/library-item.raw.css
```diff
@@ -0,0 +1,54 @@
...
+    transform: translate(0, 1.2rem);
+}
+
+.driver-popover.tooltip-face-sensing-modal .driver-popover-description {
+    font-weight: 400;
+    line-height: 1.25rem;
+    font-size: 0.875rem;
+}
+
+.driver-popover.tooltip-face-sensing-modal .driver-popover-arrow-side-right {
```
KManolov3, line 41: Is this style used?

adzhindzhi: Yep, depending on the screen size the face sensing item can be either last or first in line, so the tooltip may display both on the left and right. For example:
image
image

PR #341, opened 2025-10-02
[UEPR-385] Fix sprites issues when extension button tooltip is displayed

adzhindzhi:
### Resolves

[UEPR-385](https://scratchfoundation.atlassian.net/browse/UEPR-385)

### Proposed Changes

- Use more specific selectors to override driverjs styles

### Reason for Changes

Make sprites and other affected items clickable

[UEPR-385]: https://scratchfoundation.atlassian.net/browse/UEPR-385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/scratch-gui/src/components/alerts/alert.css
+++ b/packages/scratch-gui/src/components/alerts/alert.css
@@ -2,7 +2,7 @@
 @import "../../css/colors.css";
 @import "../../css/z-index.css";
 
-.alert {
+body .alert {
     width: 100%;
     display: flex;
     flex-direction: row;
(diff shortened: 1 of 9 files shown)


KManolov3 (approved):
A more resilient approach for new components, that ensures such behavior out of the box and also protects against naming collisions (which css modules also do in the editor, but not in www) is to qualify styles with the container they are in, where applicable. But as far as I see in this PR - that's not applicable to all cases, so I am okay as-is.

PR #360, opened 2025-10-22
feat: add onClickLogin on the main GUI component

georgyangelov:
Allows overriding what happens when the Login button in the header is clicked. In some cases we want to be able to open a popup, not just render a user/pass dialog as in current Scratch.

### Proposed Changes

Exposes `onClickLogin` event to the embedder app. Preserves current behavior if it's not passed.

### Reason for Changes

NGP

### Test Coverage

I have not ^^ but current tests should still pass.


MiroslavDionisiev (approved):

KManolov3 (approved):
Looks good and safe

PR #361, opened 2025-10-22
Update LICENSE and TRADEMARK files

### LINKED ISSUE #359: Licensing Discrepancy
It looks like each of the sub-projects/packages in this monorepo still have MIT licenses at their root (all of the copyrights haven't been updated since 2016 and the commit says it was from 9yrs ago), but the root project is AGPLv3 (its addition seemingly from a year ago coming with a commit message that says something like "switching to AGPL").

I'm assuming that the AGPLv3 is supposed to cover all the projects and the MIT licenses were just left behind, but this may be confusing as I believe some projects with complex licensing structures sometimes offer sub components of a project with more permissive licensing.

The reason why I discovered this at all was because I looked to see if the VM was maybe licensed differently than the rest of the project to make embedding it lower friction for 3rd parties, but then came to find that all the other packages had MIT licenses, and I am guessing that it's not _only_ the shell scripts that are intended to be covered by AGPL.

cwillisf:
### Resolves

- Fixes #359

### Proposed Changes

Makes our license and trademark policy consistent and available across:
- the repository
- all packages within the repository
- published packages in the npm registry

Specifically:
1. remove outdated `LICENSE` files
2. update `TRADEMARK` file with current information
3. add a script to copy `LICENSE` and `TRADEMARK` into each package/workspace
4. run that script and commit the results

Note that some commits associated with this work were accidentally committed to `develop` directly. I have adjusted the repository's GitHub settings to block myself from doing that again in the future :sweat_smile:

### Reason for Changes

Prior to today's changes:

- The `LICENSE` file in this repository's root disagreed with the `LICENSE` file in each of the workspaces/packages, which retained the old license.
- There was no `TRADEMARK` file in the root of the repository.
- The `TRADEMARK` file in each workspace/package was out of date.

### DIFF
--- a/TRADEMARK
+++ b/TRADEMARK
@@ -1,0 +1,1 @@
+The Scratch trademarks, including the Scratch name, logo, the Scratch Cat, Gobo, Pico, Nano, Tera and Giga graphics (the "Marks"), are property of the Scratch Foundation. Marks may not be used to endorse or promote products derived from this software without specific prior written permission.
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "prepare": "husky install",
     "test": "npm test --workspaces",
     "refresh-gh-workflow": "ts-node scripts/build-gha-workflows.ts",
+    "update-legal": "npm --workspaces exec -c 'rm -f ./{LICENSE,TRADEMARK} && cp -f ../../{LICENSE,TRADEMARK} .'",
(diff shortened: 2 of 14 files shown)


cwillisf:
GitHub's diffing doesn't seem to handle symlinks correctly: the LICENSE file in each package was previously a symlink, and is now a regular file containing the text of the AGPL v3.0 license.

PR #363, opened 2025-10-23
Update jest and ts-jest to v29

cwillisf:
### Resolves

- Attempts to update `scratch-paint` in `scratch-gui` are currently failing due to a dependency conflict between versions of the `typescript` package.

### Proposed Changes

Update `jest` and `ts-jest` to version 29.

Supporting change: `scratch-vm` now explicitly exports block types.

BREAKING CHANGE: `scratch-vm` now lists default exports under ".", which may be incompatible with some tools.

### Reason for Changes

The old version of `ts-jest` was only compatible with `typescript@<4`. Since `scratch-paint` indirectly requires `typescript@^5`, that was a problem.

Updating to a newer version of `ts-jest` required updating `jest` and (in our case) adding `jest-environment-jsdom`. Major version mismatches in that group can cause very strange failures, such as jestjs/jest#15674.

This caused some minor changes in snapshot behavior and async timing, so I fixed up some unit tests for that.

Updating to a newer `jest` also caused a change in module resolution, breaking the import of `scratch-vm`'s `BlockType` enum in `scratch-gui/test/unit/util/define-dynamic-block.test.js`. To fix that, I added an explicit export for `block-type` in the `exports` section of `scratch-vm`'s `package.json` file.

Adding a sub-path export to `scratch-vm`'s `exports` meant adjusting the top-level exports to all be subpaths, moving the existing conditional exports underneath the `"."` subpath.

I don't know why she swallowed the fly.

### Test Coverage

Covered by existing tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -156,7 +156,6 @@
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.28.4.tgz",
       "integrity": "sha512-2BCOP7TN8M+gVDj7/ht3hsaO/B/n5oDbiAyyvnRlNOs+u1o+JWNYTQrmpuNp1/Wq2gcFrI01JAW+paEKDMx/CA==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.3",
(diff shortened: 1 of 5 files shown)


KManolov3:
Looks good.. changes as a whole seem tame enough, given the major version bumps.
Hopefully we don't reach the latter verses of the rhyme :)

KManolov3 (approved):

PR #367, opened 2025-10-29
Merge React 18 hotfix to develop

cwillisf:
### Proposed Changes

Merge the changes from the recent React 18 hotfix to `develop`

### Reason for Changes

Don't want to un-deploy the fix with our next regular release :)

### Test Coverage

The fix itself has been through QA already. We'll manually test the integration with our upcoming QA cycle. This one's tricky to test in an automated way, especially since we actually don't want this behavior in the long run.


cwillisf:
Merging to get `develop` settled, then I'll work on the test failures. I suspect the problem is related to `package-lock.json` weirdness and not a "real" test failure, but we'll see...

PR #374, opened 2025-11-04
feat: support cat-blocks as an optional theme

KManolov3:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-416

### Proposed Changes

- Rename "theme" to "color mode" in most places in the code to better align with its current meaning.
- Extract a reusable component for persistent editor settings
- Add a new theme menu, with an option for cat-blocks

### Reason for Changes

As part of the membership initiative, we want to reintroduce the cat blocks theme as an optional theme for supporters of Scratch


adzhindzhi (commented):
Great start! It does raise some UX questions, such as how we would toggle this on/off, where it would live in the UI, and whether we'd want to use it along with the high contrast theme (likely yes), etc. But that's not something we need to figure out now.

packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -518,6 +523,8 @@ GUIComponent.defaultProps = {
     basePath: './',
     blocksId: 'original',
     canChangeLanguage: true,
+    canChangeColorMode: true,
+    // TODO: Make this depend on membership status
     canChangeTheme: true,
```
cwillisf: I don't think any change is necessary right now, but I can imagine a world where everyone can change themes, but some specific themes are exclusive. It might be that the theme menu should show up if `.some(theme => theme.isAvailable(me))` or something like that. That could work if, for example, we still want cat blocks for everyone once a year.

packages/scratch-gui/src/containers/blocks.jsx
```diff
@@ -699,7 +700,7 @@ const mapStateToProps = state => ({
     toolboxXML: state.scratchGui.toolbox.toolboxXML,
     customProceduresVisible: state.scratchGui.customProcedures.active,
     workspaceMetrics: state.scratchGui.workspaceMetrics,
-    useCatBlocks: isTimeTravel2020(state)
+    useCatBlocks: isTimeTravel2020(state) || state.scratchGui.settings.theme === CAT_BLOCKS_THEME
```
cwillisf, line 703: :+1:

packages/scratch-gui/src/lib/settings/color-mode/persistence.js
```diff
@@ -0,0 +1,47 @@
+import cookie from 'cookie';
+
+import {DEFAULT_MODE, HIGH_CONTRAST_MODE} from '.';
+
+const PREFERS_HIGH_CONTRAST_QUERY = '(prefers-contrast: more)';
+// Technically what we are persisting is the color mode, but for historical reasons,
+// we should continue using 'scratchtheme' as the cookie key.
+const COOKIE_KEY = 'scratchtheme';
```
cwillisf, line 8: :heavy_check_mark:

cwillisf: We might want to migrate this to a new key at some point, but I'd wait until we have a clearer picture of the future of themes and color modes first.

cwillisf (approved):

adzhindzhi (approved):

cwillisf:
Looks like there's a legitimate test failure related to the settings menu. I played around with it locally, and the language & color mode menus weren't consistently closing each other as they should. I think it's related to clicking the submenus rather than hovering over them.

KManolov3:
> Looks like there's a legitimate test failure related to the settings menu. I played around with it locally, and the language & color mode menus weren't consistently closing each other as they should. I think it's related to clicking the submenus rather than hovering over them.

Yeah, that was my bad, I seem to have been a bit overzealous in refactoring 😅. I've returned the corresponding redux actions ensuring only one sibling menu is open.

PR #376, opened 2025-11-07
is online?

cwillisf:
See also scratchfoundation/scratch-blocks#3412

### Proposed Changes

Add a new block to the Sensing category to detect whether or not the system is online (has an active Internet connection).

### Reason for Changes

Scratch users have discovered ways to check for online status, most commonly by repeatedly using the Translate extension. While I applaud the ingenuity, this approach generates excess load on the Translate infrastructure. Scratchers and the Scratch Team will all have a better time if we provide a direct way to check for online status.

I'm open to feedback, including:
- Should the label of the monitor be something other than `online`?
- I definitely think this block belongs in the Sensing category, but is this (just before username) the right place for it within the category?

### Test Coverage

Tested locally:
image

### DIFF
--- a/packages/scratch-gui/src/lib/make-toolbox-xml.js
+++ b/packages/scratch-gui/src/lib/make-toolbox-xml.js
@@ -5,7 +5,7 @@ const categorySeparator = '<sep gap="36"/>';
 
 const blockSeparator = '<sep gap="36"/>'; // At default scale, about 28px
 
- 
+
 const motion = function (isInitialSetup, isStage, targetId, colors) {
     const stageSelected = ScratchBlocks.ScratchMsgs.translate(
         'MOTION_STAGE_SELECTED',
@@ -513,6 +513,7 @@ const sensing = function (isInitialSetup, isStage, targetId, colors) {
         <block id="current" type="sensing_current"/>
         <block type="sensing_dayssince2000"/>
         ${blockSeparator}
+        <block id="online" type="sensing_online" />
         <block type="sensing_username"/>
         ${categorySeparator}
     </category>
@@ -736,7 +737,7 @@ const myBlocks = function (isInitialSetup, isStage, targetId, colors) {
     </category>
     `;
 };
- 
+
 
 const xmlOpen = '<xml style="display: none">';
 const xmlClose = '</xml>';
--- a/packages/scratch-gui/src/lib/opcode-labels.js
+++ b/packages/scratch-gui/src/lib/opcode-labels.js
@@ -68,6 +68,11 @@ const messages = defineMessages({
         description: 'Label for the loudness monitor when shown on the stage',
         id: 'gui.opcodeLabels.loudness'
     },
+    sensing_online: {
+        defaultMessage: 'online',
+        description: 'Label for the online status monitor when shown on the stage',
+        id: 'gui.opcodeLabels.online'
+    },
     sensing_username: {
         defaultMessage: 'username',
         description: 'Label for the username monitor when shown on the stage',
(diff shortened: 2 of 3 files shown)


CST1229:
> Should the label of the monitor be something other than online?

Honestly, I thought block monitors were always named directly after the block (so, `is online?`). But since that's not the case, `online` is also fine.

ampelectrecuted:
FINALLY a new block!

redspacecat:
This would be amazing! Mostly just because it's the first new block in a while (besides face sensing) but also it would be pretty useful.

KManolov3 (approved):
Looks good to me. Congrats on a new block! 🎉

ericrosenbaum:
I recommend changing both the wording on the block and in the monitor to just `online?`

cwillisf:
@ericrosenbaum, I've renamed the block to `online?`. That was a change in `scratch-blocks`, so it's not obvious here.

mrsrec:
> @ericrosenbaum, I've renamed the block to `online?`. That was a change in `scratch-blocks`, so it's not obvious here.

I still think the name is confusing. "online?" could mean whether the project has internet connectivity functionality, whether or not it's actively using them at that moment, or whether it's being run in the online editor vs. the offline editor. Something like "connected to internet?" would solve the third one, but not the first two. Maybe "internet available?", but I'm not sure.

The other thing is whether or not a block like this makes sense in the sensing category, as opposed to one of the extensions. To me it seems a bit out of place (generally anything relating to the internet is abstracted away there), but I can't think of any better place to put it, and an extension that only has one block wouldn't make sense.

mrsrec:
You did say in the OP you were open to feedback, but IDK if it' too late for that now.

CST1229:
> I still think the name is confusing. "online?" could mean whether the project has internet connectivity functionality, whether or not it's actively using them at that moment, or whether it's being run in the online editor vs. the offline editor. Something like "connected to internet?" would solve the third one, but not the first two. Maybe "internet available?", but I'm not sure.

there's also another possible interpretation, "is [a user, possibly the creator of the project] currently online on the Scratch website?" (so like, a person might try to use the "is online?" block to make a project that shows if they're online on the website or not). though "online?" does probably have a lower chance of users interpreting it that way than "is online?", and the two names you suggested don't have that possibility at all

ampelectrecuted:
also the old name certainly rings a bell to a certain browser extension from 2017

SpinningCube:
If I understand correctly, this block would be most useful for projects using Translate or Text to Speech (and potentially all the Scratch Link extensions but I don't understand how those work). I see similar functionality being potentially useful for cloud variables as well, but for cloud variables one would also want to check whether the cloud variables are online / the user is a new scratcher. But I wonder if it would be more useful for each extension to have its own boolean for availability (cloud variables too), in the albeit unlikely case that a specific service is offline.

rosics-code:
@cwillisf What if we could also detect if a user is online on Scratch?

CST1229:
> What if we could also detect if a user is online on Scratch?

isn't that a rejected suggestion
<sup>(self-marked as off-topic)</sup>

rosics-code:
> > What if we could also detect if a user is online on Scratch?
> 
> isn't that a rejected suggestion (self-marked as off-topic)

I mean, it’s not like rejected suggestions are a stop for these types of suggestions. ST can un reject a suggestion.

redspacecat:
Is this going to be added sometime?

cwillisf:
Thanks for the feedback, everyone! I'm moving forward with this as is, though there are a few more steps before it's released, so it's still possible there will be further changes.

PR #393, opened 2025-12-04
Add QueueManager & improve TaskQueue

### LINKED ISSUE #7111: CE-132 Loading a project with many assets can fail
### Expected Behavior

Any given project should either load correctly or should display a Scratch error messages. Repeated attempts to load the same project should be consistent.

### Actual Behavior

If a project has many assets (>2000) it can cause bad behavior. For example:
- the project might fail to load with a Scratch bluescreen on the first attempt, but succeed on a later attempt (once some of the assets are cached)
  - This has been observed on the website when some assets fail to load with `net::ERR_INSUFFICIENT_RESOURCES` in the network tab.
- the project might crash Chrome / Chromium's rendering process
  - This has been observed in the desktop app. From a user perspective, the whole editor is replaced by a blank white window.

See also FreshDesk ticket 323025, especially the "Technical Issue Link" attached to that ticket.

### Steps to Reproduce

1. Find or create a project with >2000 assets
   - The example I have on hand has 2114 assets. It's possible the threshold is much lower; I haven't tested.
2. Save the project on the website, or to an SB3 file
3. Try to load the project

### Operating System and Browser

White screen: Scratch desktop app on Windows
Other errors: Chrome / Chromium, any OS

cwillisf:
### Resolves

Supports an upcoming `scratch-storage` change that will resolve scratchfoundation/scratch-gui#7111 and POD-238

### Proposed Changes

Main change: in `task-herder`, add `QueueManager`, a way to conveniently manage multiple throttling queues.

Supporting changes:
- Add tests for the new `QueueManager`
- Fix and simplify the concurrency logic in `TaskQueue`
- Support vendor chunks in `scratch-storage`, which helps when using `npm link scratch-storage`
- Add `{willReadFrequently: true}` to the `Canvas.getContext('2d')` call in the bitmap load path

### Reason for Changes

- The upcoming `scratch-storage` change will use this new `QueueManager` to manage per-host throttling queues.
- The previous concurrency logic didn't provide back pressure from the concurrency limiter
  - Without that back pressure, the client needs to make extra effort to avoid growing the queue unbounded.
  - It's also not consistent with the rate-limiting behavior.
  - Removing the `p-limit` dependency is a nice side effect. (There's nothing wrong with `p-limit`, but fewer dependencies is better.)
- The `willReadFrequently` change:
  - Matches our actual intent with the canvas/context
  - Fixes a Chrome console warning
  - Fixes a Chrome bug(?) that corrupted buffers when the asset download allowed concurrency > 5
  - Might improve performance

### Test Coverage

The new `QueueManager` is covered by new tests. Existing tests cover everything else.


KManolov3 (approved):
This looks good to me! Thank you for adding tests!

PR #394, opened 2025-12-09
[UEPR-336] Implement Semantic Layout in the Editor

kbangelov:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-336

### Proposed Changes

I have added role and aria-label tags for the elements to be considered as landmarks and be navigated through with a screen reader easily. I have separated the page into three parts : 
- Menu topbar (role = banner)
- editor (role = main)
- Stage and Target (role = complementary)

Those three I have separated into the following:
Editor:
- TabList (role = region)
- Code/(Costumes/Backdrops)/Sound Editor Panel (role="region")
- Backpack (region)

Stage and Target:
- Stage (region)
- Target (region)

### Reason for Changes

None of them were focusable so far.

### To be tested with a proper screen reader but it seems to work
I have not fully tested whether they can be navigated through quickly as landmarks yet, because the only semi-adequate screen reader I found was an extension which doesn't let me tab from landmark to landmark but reads them out loud when hovered. Also another extension for landmarks does find all 8 described above. So I have tested it to the best of my abilities, still a check from a proper screen will be much appreciated. I think it works.


packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -403,7 +409,11 @@ const GUIComponent = props => {
                             ) : null}
                         </Box>
 
-                        <Box className={classNames(styles.stageAndTargetWrapper, styles[stageSize])}>
+                        <Box
+                            aria-label="Stage and Target"
+                            role="complementary"
+                            className={classNames(styles.stageAndTargetWrapper, styles[stageSize])}
+                        >
                             <StageWrapper
```
KManolov3: I think both the stage and target should be considered sections on their own.

packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -253,6 +253,8 @@ const GUIComponent = props => {
                     />
                 ) : null}
                 {!menuBarHidden && <MenuBar
+                    role="navigation"
```
KManolov3: I think banner role might make the most sense here. 
Maybe the label should also become something like "Menu topbar"

packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -290,14 +292,20 @@ const GUIComponent = props => {
                 />}
                 <Box className={boxStyles}>
                     <Box className={styles.flexWrapper}>
```
KManolov3: The `boxStyles` Box has only one child element - the `flexWrapper`. I imagine that might create unnecessary overhead for screen reader. We'd want to either set ariaHidden to one (or both) of them, or merge them together. Can you try merging the two Box-es into one (passing all of their styles)? If it results in unintended consequences for the layout, we can go with another approach.

packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -399,19 +413,33 @@ const GUIComponent = props => {
                                 </TabPanel>
                             </Tabs>
                             {backpackVisible ? (
-                                <Backpack host={backpackHost} />
+                                <Backpack
+                                    host={backpackHost}
+                                    role="region"
+                                    ariaLabel="Backpack"
```
KManolov3: It's unfortunate that we end up passing "aria-label" (kebab case) vs ariaLabel (camel case) in different places, as that creates a lot of potential for developer error, but I understand that PropTypes is somewhat forcing our hand on this. We can possibly return to this again later.

packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -310,7 +318,10 @@ const GUIComponent = props => {
                                 // Actual behavior: the Code or Sounds tab is now focused
                                 focusTabOnClick={false}
                             >
-                                <TabList className={tabClassNames.tabList}>
+                                <TabList
+                                    className={tabClassNames.tabList}
+                                    aria-roledescription="tab list"
```
KManolov3: The `react-tabs` library already applies a tablist role, so this description might be redundant

packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -359,7 +370,10 @@ const GUIComponent = props => {
                                         />
                                     </Tab>
                                 </TabList>
-                                <TabPanel className={tabClassNames.tabPanel}>
+                                <TabPanel
+                                    className={tabClassNames.tabPanel}
+                                    aria-roledescription="editor panel"
```
KManolov3: Let's use `aria-label` here. I think it's useful to have these labeled, as they serve a semantic purpose, but not sure if `aria-roledescription` is the correct property here

KManolov3: Additionally, what do you think about moving this aria-label to the tabs themselves:
- "Code Editor Tab"
- "Costumes Tab"
- "Sounds Tab"
?

packages/scratch-gui/src/components/menu-bar/menu-bar.jsx
```diff
@@ -437,6 +437,8 @@ class MenuBar extends React.Component {
                     this.props.className,
                     styles.menuBar
                 )}
+                aria-label={this.props.ariaLabel}
+                role="navigation"
```
KManolov3: Aren't we passing the role from outside?

packages/scratch-gui/src/components/menu-bar/menu-bar.jsx
```diff
@@ -971,6 +974,7 @@ MenuBar.propTypes = {
     platform: PropTypes.oneOf(Object.keys(PLATFORM)),
     projectTitle: PropTypes.string,
     renderLogin: PropTypes.func,
+    role: PropTypes.string,
```
KManolov3: Nitpick: If we are defining custom aria properties for some of these Wrapper components, I think `ariaRole` might make their purpose clearer.

packages/scratch-gui/src/components/gui/gui.jsx
kbangelov: I am leaving an aria label, let me know whether you think it's appropriate to have "tablist (role) editor tabs (aria-label)"

packages/scratch-gui/src/components/gui/gui.jsx
kbangelov: good idea

packages/scratch-gui/src/components/menu-bar/menu-bar.jsx
kbangelov: smh

packages/scratch-gui/src/components/menu-bar/menu-bar.jsx
kbangelov: yeah, fair

packages/scratch-gui/src/components/gui/gui.jsx
kbangelov: Good observation, I believe I have merged them without any visual changes, I only had to remove a repeating "height" value.

packages/scratch-gui/src/components/backpack/backpack.jsx
```diff
@@ -53,11 +54,16 @@ const Backpack = ({
     onDelete,
     onMouseEnter,
     onMouseLeave,
-    onMore
+    onMore,
+    ariaRole
```
KManolov3: Nitpick: Let's sort properties alphabetically when possible for readibility

packages/scratch-gui/src/components/gui/gui.css
```diff
@@ -28,7 +28,6 @@
         b) for combined stage menu + stage + sprite/stage selectors
     */
     flex-direction: row;
-    height: 100%;
```
KManolov3, line 31: I imagine this was no longer needed because of the `body-wrapper` css class?

packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -288,137 +290,166 @@ const GUIComponent = props => {
...
                             >
-                                <TabList className={tabClassNames.tabList}>
-                                    <Tab className={tabClassNames.tab}>
-                                        <img
-                                            draggable={false}
-                                            src={codeIcon}
-                                        />
+                                <Tab
+                                    className={tabClassNames.tab}
+                                    aria-label="Code Editor Tab"
```
KManolov3: I rather meant for the aria-labels to be on the components rendering the tabs (so the <Box wrapping the Blocks, the <CostumeTab component, the <SoundTab component). I see that I wasn't very clear what I meant in my previous comment, sorry about that.

In the case of the <Tab components, we shouldn't have an issue there, as the messages describing them (e.g. the FormattedMessage below translates into a <span> Code </span>, should make their purpose clear.

packages/scratch-gui/src/components/stage-wrapper/stage-wrapper.jsx
```diff
@@ -13,12 +13,14 @@ import styles from './stage-wrapper.css';
...
 const StageWrapperComponent = function (props) {
     const {
+        ariaLabel,
         isFullScreen,
         isRtl,
         isRendererSupported,
         loading,
         manuallySaveThumbnails,
         onUpdateProjectThumbnail,
+        ariaRole,
```
KManolov3: nitpick: see comment above about alphabetical prop ordering

packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -290,168 +290,166 @@ const GUIComponent = props => {
...
+                            ariaRole="region"
+                            ariaLabel="Stage"
+                        />
                         <Box
-                            role="complementary"
-                            aria-label="Stage and Target"
-                            className={classNames(styles.stageAndTargetWrapper, styles[stageSize])}
+                            className={styles.targetWrapper}
+                            role="region"
+                            aria-label="Target"
```
KManolov3: Nitpick: "Target" -> "Target Pane"

packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -290,168 +290,166 @@ const GUIComponent = props => {
...
+                        className={styles.editorWrapper}
+                    >
+                        <Tabs
+                            forceRenderTabPanel
+                            className={tabClassNames.tabs}
+                            selectedIndex={activeTabIndex}
+                            selectedTabClassName={tabClassNames.tabSelected}
+                            selectedTabPanelClassName={tabClassNames.tabPanelSelected}
+                            onSelect={onActivateTab}
+                            role="region"
```
KManolov3: Having the element directly under the "main" landmark be a separate region is probably redundant.

Instead, I am split between either not having any regions, besides the backpack in it, or:
- the <TabList becomes a region
- the rendered tab content (the <Box wrapping the Blocks, the <CostumeTab component, the <SoundTab component) also become regions - and only the active one will be rendered (and so selectable) at a single time

As I don't have a strong argument against doing it, I think the latter option (defining additional regions) should be the one to go with.

packages/scratch-gui/src/components/backpack/backpack.jsx
kbangelov: Yeah, I clearly forgot about it while changing the name

packages/scratch-gui/src/components/gui/gui.css
kbangelov: Yes, that class being on the inside used to say "make up 100% of the body wrapper height"

packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -359,66 +368,94 @@ const GUIComponent = props => {
...
-                                            username={username}
-                                        />
-                                    </Box>
-                                    <ExtensionsButton
-                                        activeTabIndex={activeTabIndex}
-                                        intl={intl}
+                            </Box>
+                            <TabPanel
+                                className={tabClassNames.tabPanel}
+                                aria-label="Code Editor Panel"
```
KManolov3: Can we remove the aria-label here and leave it only on the region to avoid confusion?

KManolov3 (approved):

packages/scratch-gui/src/components/gui/gui.jsx
kbangelov: Absolutely

KManolov3:
FYI @cwillisf

PR #405, opened 2025-12-30
[UEPR-454] Add translations to "aria-label"s in editor

kbangelov:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-454

### Proposed Changes

added **intl.formatMessage** logic

### DIFF
--- a/packages/scratch-gui/src/components/gui/gui.jsx
+++ b/packages/scratch-gui/src/components/gui/gui.jsx
@@ -2,7 +2,7 @@ import classNames from 'classnames';
 import omit from 'lodash.omit';
 import PropTypes from 'prop-types';
 import React, {useEffect, useCallback} from 'react';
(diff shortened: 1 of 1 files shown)


kbangelov:
The aria-labels don't seem to change after language switch, should be tested with a real screen reader.

KManolov3:
@kbangelov that makes sense, since they were just added as translation "objects" (and are not even yet pushed to Transifex -> they are not translated there) - hence the code has no reference point for those strings to draw from for other languages and it defaults to English. We can test that the translation *logic* works by manually modifying the translation files generated by scratch-l10n.

KManolov3 (approved):
We should test the translated labels get picked up by a screen reader, but otherwise looks good to me 🙌

adzhindzhi (approved):

kbangelov:
I tested with custom german translations and it seemed to work fine.

PR #408, opened 2026-01-05
[UEPR-342] Refactor editor components to use semantic HTML

adzhindzhi:
### Resolves

[UEPR-342](https://scratchfoundation.atlassian.net/browse/UEPR-342)

### Proposed Changes

- Use `<button>` element, instead of div/span where applicable
- Use semantic lists on the main editor page where applicable 
- Use semantic page layout where applicable

TODOs:
- Use semantic components in the navbar - will be handled when ensuring the navbar is navigable (or in a separate task if it requires too much changes)

### Reason for Changes

- This task is part of a larger ongoing effort to make the editor more accessible 

### Test Coverage

- Updated tests to comply with the changes

[UEPR-342]: https://scratchfoundation.atlassian.net/browse/UEPR-342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/scratch-gui/src/components/sprite-selector-item/sprite-selector-item.jsx
```diff
@@ -28,6 +28,7 @@ const SpriteSelectorItem = props => {
                 disabled={props.preventContextMenu}
                 asChild
             >
+                {/* Does it make sense for this to be a button */}
```
cwillisf: I did some reading just now, and I think `role="grid"` might be the best bet for the sprite selector, with `gridcell` for individual sprite tiles. Double-check me on that, of course :sweat_smile:

cwillisf (approved):
Looks great so far!

packages/scratch-gui/src/components/delete-button/delete-button.jsx
```diff
@@ -6,13 +6,13 @@ import styles from './delete-button.css';
 import deleteIcon from './icon--delete.svg';
 
 const DeleteButton = props => (
-    <div
+    // TODO: Make the focus visible
```
kbangelov: what do you mean by that? I think it already is focusable via tab navigation? There is some highlight already too. I don't know if it's perfect as styling, seems debatable, but it is there.

packages/scratch-gui/src/components/extension-button/extension-button.jsx
```diff
@@ -140,6 +140,7 @@ const ExtensionButton = props => {
 
     return (
         <Box className={styles.extensionButtonContainer}>
+            {/* TODO: Add focus indicator */}
```
kbangelov, line 143: I'm guessing you mean you want it to be highlighted better than it currently is?

packages/scratch-gui/src/components/green-flag/green-flag.jsx
```diff
@@ -14,20 +14,24 @@ const GreenFlagComponent = function (props) {
...
+                    className,
+                    styles.greenFlag,
+                    {
+                        [styles.isActive]: active
+                    }
+                )}
+                draggable={false}
+                src={greenFlagIcon}
+                title={title}
+                {...componentProps}
```
kbangelov: Since this used to be an image with onClick, where do you think the **...componentProps** should go? The button or the image? I think from a certain POV it makes sense for it to go to the top element in the return, since a button is what the element is. Currently we don't utilize the extra props anyways. Sure, we have an **active** prop, which indirectly changes the styling. And it seems more related to the image. But I'd still say it should go to button.

Since it doesn't matter for now, I'm fine with leaving it as it is.

packages/scratch-gui/src/components/stop-all/stop-all.jsx
```diff
@@ -14,20 +14,24 @@ const StopAllComponent = function (props) {
...
+                    className,
+                    styles.stopAll,
+                    {
+                        [styles.isActive]: active
+                    }
+                )}
+                draggable={false}
+                src={stopAllIcon}
+                title={title}
+                {...componentProps}
```
kbangelov: same comment here

packages/scratch-gui/src/components/sound-editor/sound-editor.jsx
```diff
@@ -149,6 +149,7 @@ const SoundEditor = props => {
                 <div className={styles.inputGroup}>
                     <Label text={intl.formatMessage(messages.sound)}>
                         <BufferedInput
+                            // TODO: Why 1? Update when working on correct order of navigation
```
kbangelov, line 152: I'm guessing the logic is "go to that input element first when you click tab after opening the sound editor". But there's a bug, because it doesn't do that. What it does is that it makes it the first focusable item on the entire page, before it goes to nav bar elements or anything else. Which is not what you want at all. Which is why **tabIndex={1}** is generally not recommended. Changing it to 0 makes a whole lot of sense for me except for a very weird bug I observed.

If you change it to 0 and get to the sound editor panel just via tab navigation before clicking ANYTHING else on the page, you get some errors that I don't feel like following in their entirety right now. Leaving that comment seems like the correct approach since it's probably for another PR. Perhaps you can leave a comment for this bug I just found as well (as an extension to the comment you already left)

packages/scratch-gui/src/components/sprite-selector-item/sprite-selector-item.jsx
kbangelov: I think yes? Just intuitively having onClick is very button-like behaviour. You eventually want tab navigation for the sprites as well. Having both **onClick** and **onMouseEnter** is a little weird to me, even after following the logic backwards for that method, I am not sure what the aim is. They are not focusable yet so it's not easy to test it either. I'll look into that in my PR.

kbangelov (commented):
Good stuff. Left a few comments. Please address when you are able to.

packages/scratch-gui/src/components/delete-button/delete-button.jsx
adzhindzhi: Yep, you're right, that comment is no longer valid.

packages/scratch-gui/src/components/extension-button/extension-button.jsx
adzhindzhi: Overall, yes. I left it as a comment so that we'd remember to check it out with [UEPR-339](https://scratchfoundation.atlassian.net/browse/UEPR-339)

packages/scratch-gui/src/components/green-flag/green-flag.jsx
adzhindzhi: Hm, fair question, I agree it makes more sense for them to be passed to the top component.

packages/scratch-gui/src/components/sound-editor/sound-editor.jsx
adzhindzhi: Yeah, overall we have a task for ensuring the correct order of navigation - [UEPR-344](https://scratchfoundation.atlassian.net/browse/UEPR-344), I left this TODO as a reminder for when we start working on it.

packages/scratch-gui/src/components/sprite-selector-item/sprite-selector-item.jsx
adzhindzhi: This is exactly where I was coming from with my original comment - the card should behave like a button because it’s clickable. The tricky part is that it has two actions: selecting the costume and deleting it via the `DeleteButton` (when selected). Making the outer element a real `<button>` would be ideal for semantics, but it would lead to invalid HTML unless we split the card into two buttons. Considering accessibility, keyboard support, and valid markup, I think the best approach is a `<div role="button" tabIndex={0}>` (or `grid/gridcell` as Chris suggested). This keeps it focusable and "button-like" for selection while allowing the `DeleteButton` to remain valid. We’d have to handle `onKeyDown` manually to make it fully interactive, but that seems fine to me.

adzhindzhi: I’m honestly a little torn between using `role="button"` for each card and going with `grid/gridcell`. Both seem fine to me 😅 `role="button"` is simple and accessible, but using `grid/gridcell` could make keyboard navigation a bit more structured if we wanted to take it further.

packages/scratch-gui/src/components/asset-panel/asset-panel.jsx
```diff
@@ -12,12 +12,16 @@ const AssetPanel = props => {
...
+        element="section"
     >
         <Selector
             className={styles.selector}
             {...restProps}
         />
-        <Box className={styles.detailArea}>
+        <Box
+            className={styles.detailArea}
+            element="section"
```
KManolov3, line 23: Do we want to have both this and its container be < section > elements? Maybe just the container?

KManolov3 (approved):
Looks good to me 🙌

packages/scratch-gui/src/components/sprite-selector-item/sprite-selector-item.jsx
kbangelov: Agreed

kbangelov (approved):

packages/scratch-gui/src/components/asset-panel/asset-panel.jsx
adzhindzhi: My thinking was that the tab content itself makes sense as a separate `<section>`, since it represents a distinct part of the page. Within it, there's an `<aside>` element for the asset list (sounds or costumes) and another `<section>` for the actual paint/sound editor. Conceptually, it made sense to me to treat both as sections, but I’m definitely open to changing this if you think it'd improve the structure.

PR #411, opened 2026-01-09
[UEPR-470] Fix misaligned cat ears frame on some user avatars

adzhindzhi:
### Resolves

- Resolves [UEPR-470](https://scratchfoundation.atlassian.net/browse/UEPR-470)

### Proposed Changes

- Move the margins to a wrapper elements, rather than duplicating them both on the actual image and the badge wrapper

### Reason for Changes

- Member avatar images can appear misaligned when the image has a narrow width. The issue was caused by CSS margins applied to both the avatar wrapper and the `img` element. When the image is smaller, the margin forces the image to shrink instead of filling the fixed avatar frame, resulting in a visible gap.

[UEPR-470]: https://scratchfoundation.atlassian.net/browse/UEPR-470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/scratch-gui/src/components/menu-bar/account-nav.css
+++ b/packages/scratch-gui/src/components/menu-bar/account-nav.css
@@ -19,11 +19,11 @@
     font-weight: normal;
 }
 
-[dir="ltr"] .user-info .avatar {
+[dir="ltr"] .user-info .avatar-wrapper {
     margin-right: calc($space * .8125);
 }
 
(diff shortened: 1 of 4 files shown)


KManolov3 (approved):
Looks good!

cwillisf (approved):

PR #413, opened 2026-01-12
fix: improve support for isolatedModules flag

pierluca:
### Resolves

As part of our development, we made use of the TypeScript [isolatedModules flag](https://www.typescriptlang.org/tsconfig/#isolatedModules) and this raised an issue when compiling the project along the lines of:
> Re-exporting a type when the '--isolatedModules' flag is provided requires using 'export type'.ts

### Proposed Changes

This PR implements the `export type` changes as discussed in [this Stackoverflow thread](https://stackoverflow.com/questions/70416680/re-exporting-a-type-when-the-isolatedmodules-flag-is-provided-requires-using), as well as a minor JSDoc fix.

### Reason for Changes

These changes are a tiny step in the direction of supporting `isolatedModules` in Scratch and they improve correctness.

### Test Coverage

I'm not sure what would make sense here, but I'm open to suggestions

### DIFF
--- a/packages/scratch-gui/src/index-standalone.tsx
+++ b/packages/scratch-gui/src/index-standalone.tsx
@@ -6,8 +6,8 @@ import {EditorState} from './lib/editor-state';
 import {ReactComponentLike} from 'prop-types';
 import {compose} from 'redux';
 
-export {EditorState, EditorStateParams} from './lib/editor-state';
-export {AccountMenuOptions} from './lib/account-menu-options';
+export {EditorState, type EditorStateParams} from './lib/editor-state';
+export {type AccountMenuOptions} from './lib/account-menu-options';
 
 export {setAppElement} from 'react-modal';
 
--- a/packages/scratch-gui/src/lib/legacy-storage.ts
+++ b/packages/scratch-gui/src/lib/legacy-storage.ts
@@ -69,11 +69,13 @@ export class LegacyStorage implements GUIStorage {
         vmState: string,
         params: { originalId: string; isCopy: boolean; isRemix: boolean; title: string; }
     ): Promise<{ id: string | number; }> {
-        if (!this.projectHost) {
+        const host = this.projectHost;
+
+        if (!host) {
             return Promise.reject(new Error('Project host not set'));
         }
         // Haven't inlined the code here so that we can keep Git history on the implementation, just in case
-        return saveProjectToServer(this.projectHost, projectId, vmState, params);
+        return saveProjectToServer(host, projectId, vmState, params);
     }
 
     private cacheDefaultProject (storage: ScratchStorage) {


pierluca:

Tyratox:

KManolov3 (approved):
Looks good to me!

cc: @cwillisf @georgyangelov in case you have an opinion on this

packages/scratch-gui/src/lib/save-project-to-server.js
```diff
@@ -4,7 +4,7 @@ import xhr from 'xhr';
 /**
  * Save a project JSON to the project server.
  * This should eventually live in scratch-www.
- * @param {string} projectHost the hostname of the project service.
+ * @param {string|undefined} projectHost the hostname of the project service.
```
georgyangelov: Is it truly possible to have this as `undefined`? Wouldn't that result in ``` `${projectHost}/${qs}` ``` being `undefined/...` below?

georgyangelov (approved):
Export looks good, I left a question for the JSDoc type change

packages/scratch-gui/src/lib/save-project-to-server.js
Tyratox: It probably should never be, I agree but the typing in

https://github.com/scratchfoundation/scratch-editor/blob/f7e83ec7c5a8f07104c2c1ae2790003738e484ac/packages/scratch-gui/src/lib/legacy-storage.ts#L6-L9

with callsite

https://github.com/scratchfoundation/scratch-editor/blob/f7e83ec7c5a8f07104c2c1ae2790003738e484ac/packages/scratch-gui/src/lib/legacy-storage.ts#L76

is such that it may pass `undefined`. I agree that it would be good to also fix this there but this small change is effectively just making sure that the typing in the different files is compatible. Depending on the typescript compiler, the build will otherwise file since jsdoc is used to infer type information in js files: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html

georgyangelov: @Tyratox the line that you link in legacy-storage.ts performs a check and only calls `saveProjectToServer` if the projectHost is defined. Or that's the intention at least.. given that it's using `this.projectHost` in both places I see how some TS version can consider it a different check and not narrow the type.

Screenshot 2026-01-19 at 11 51 39

I suggest instead of allowing undefined as a parameter, change the `saveProject` function to:

```ts
const host = this.projectHost;

if (!host) {
    return Promise.reject(new Error('Project host not set'));
}
// Haven't inlined the code here so that we can keep Git history on the implementation, just in case
return saveProjectToServer(host, projectId, vmState, params);
```

That way TS will 100% know that the `host` variable cannot be undefined at the time where saveProjectToServer is called.

I believe that is the only call to the saveProjectToServer

Tyratox: Sounds good, happy to modify the PR like that. It is the much better solution if we know that it will only ever be called with a defined value (including potential future callsites)! Thanks :)

PR #414, opened 2026-01-13
Update Amon in sprite and costume libraries

ericrosenbaum:


KManolov3 (approved):
Looks good to me :) Ayshe released a [patch version of the editor](https://github.com/scratchfoundation/scratch-editor/releases/tag/v12.1.5), so this can go out with the nearest -www release, feel free to also merge it in develop.

ericrosenbaum:
I just want to note here for future reference (especially @cwillisf ):
- The asset id used here is not associated with a project owned by the Scratch account we usually use for that purpose (hopefully not an issue - only a concern in the event of an orphaned asset purge in the future? not sure if that would affect it either).
- I did not update the list of project ids in the scratch-libgen repo, which in the past was used to generate the library json files (I did a manual edit this time). I'm guessing that the libraries themselves will serve as source of truth for library assets in a future transition to a new system - but let me know if it would be helpful to update libgen anyway.

PR #403, opened 2025-12-19
[UEPR-445] Ensure topbar is navigable via tab navigation

kbangelov:
### Aims to resolve UEPR-445

https://scratchfoundation.atlassian.net/browse/UEPR-445

### Proposed Changes

1. Added **tabIndex=0** to the elements in nav bar so they can be focused with tab
2. Made a context (**menu-ref-context**) that takes care of tracking which menus and submenus are currently open via their refs. All of the logic with the "**isOpen**" type of props has been replaced for these menus.
3. Made a **use-menu-navigation** hook to be used by the menus. Handles common key pressing, opening, closing logic and more. Uses customized BFS logic for HTML components

### To be done additionally concerning nav bar accessibility - in other PRs?
- Choosing a **language** option does not focus on **Settings** after. Also not for the other options either, but that's because not closing after selecting an option has been their behavior so far. Perhaps another issue could be made that clears whether they should or not.
- Login Menu doesn't navigate properly (close on tab for instance) because it's written in www
- **WARNING** Because of the improper structure of nesting menu inside button we have a button inside of button with **Settings Menu** and **Language Menu / Preference Menu**, causing a warning in console.
- Focus is not too pretty on some menu items.
- Write tests for navigation bar

### To be discussed additionally
- The behavior of using screen reader and clicking with cursor leads to improper reads, as opposed to just navigating with keyboard. But I don't know if that's a big issue.

PR #416, opened 2026-01-14
[UEPR-471] Ensure Library modals are navigable through tab navigation

adzhindzhi:
### Resolves

[UEPR-471](https://scratchfoundation.atlassian.net/browse/UEPR-471)

### Proposed Changes

- Make all elements in the library modal navigable through `Tab` navigation
- Allow keyboard actions with the LibraryItems
- Add ModalFocusContext for restoring the focus to the last focused element when navigating out of the modal

TODOs:
- Feature callouts interfere with keyboard navigation in the Extensions Modal
- Tips Library is currently not navigable until #403 is merged, so restoring focus there would need to be handled separately

### Reason for Changes

- This PR is part of a larger initiative to make the Scratch Editor more accessible and navigable through keyboard.

[UEPR-471]: https://scratchfoundation.atlassian.net/browse/UEPR-471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/scratch-gui/src/components/extension-button/extension-button.jsx
+++ b/packages/scratch-gui/src/components/extension-button/extension-button.jsx
@@ -74,6 +74,10 @@ const ExtensionButton = props => {
         if (!shouldShowFaceSensingCallouts) return;
 
         const onFirstInteraction = e => {
+            // Make sure to clean up event listeners after first interaction
+            window.removeEventListener('click', onFirstInteraction);
+            window.removeEventListener('keydown', onFirstInteraction);
+
             const isExtensionButtonVisible = document.querySelector('div[class*="extension-button-container"]');
             if (!isExtensionButtonVisible) return;
 
@@ -102,10 +106,6 @@ const ExtensionButton = props => {
             setClicked(true);
             driverRef.current = tooltip;
             tooltip.drive();
-
-            // Make sure to clean up event listeners after first interaction
-            window.removeEventListener('click', onFirstInteraction);
-            window.removeEventListener('keydown', onFirstInteraction);
         };
 
         window.addEventListener('click', onFirstInteraction, {once: true});
--- a/packages/scratch-gui/src/components/extension-button/extension-button.jsx
+++ b/packages/scratch-gui/src/components/extension-button/extension-button.jsx
@@ -73,9 +73,14 @@ const ExtensionButton = props => {
     useEffect(() => {
         if (!shouldShowFaceSensingCallouts) return;
 
-        const onFirstClick = () => {
+        const onFirstInteraction = e => {
             const isExtensionButtonVisible = document.querySelector('div[class*="extension-button-container"]');
             if (!isExtensionButtonVisible) return;
+
+            if (e.type === 'keydown') {
+                // Prevent focus from jumping to the next element in the tab order after keydown finishes
+                e.preventDefault();
+            }
 
             const tooltip = driver({
                 allowClose: false,
@@ -97,8 +102,14 @@ const ExtensionButton = props => {
             setClicked(true);
             driverRef.current = tooltip;
             tooltip.drive();
+
+            // Make sure to clean up event listeners after first interaction
+            window.removeEventListener('click', onFirstInteraction);
+            window.removeEventListener('keydown', onFirstInteraction);
         };
-        window.addEventListener('click', onFirstClick, {once: true});
+
+        window.addEventListener('click', onFirstInteraction, {once: true});
+        window.addEventListener('keydown', onFirstInteraction, {once: true});
 
         return () => {
             if (driverRef.current) {
@@ -145,7 +156,6 @@ const ExtensionButton = props => {
 
     return (
         <Box className={styles.extensionButtonContainer}>
-            {/* TODO: Add focus indicator */}
             <button
                 className={
                     classNames(styles.extensionButton,
@@ -153,6 +163,7 @@ const ExtensionButton = props => {
                     )}
                 title={intl.formatMessage(messages.addExtension)}
                 onClick={handleExtensionButtonClick}
+                aria-label={intl.formatMessage(messages.addExtension)}
             >
                 <img
                     className={styles.extensionButtonIcon}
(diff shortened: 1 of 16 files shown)


kbangelov:
I like the UX

packages/scratch-gui/src/components/library-item/library-item.jsx
```diff
@@ -58,7 +58,14 @@ class LibraryItemComponent extends React.PureComponent {
                     this.props.hidden ? styles.hidden : null,
                     this.props.showItemCallout ? styles.radiate : null
                 )}
+                role="button"
```
kbangelov: could be a button?

packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -271,300 +272,302 @@ const GUIComponent = props => {
...
-                                aria-label={intl.formatMessage(ariaMessages.tabList)}
+                            <Tabs
+                                forceRenderTabPanel
+                                className={tabClassNames.tabs}
+                                selectedIndex={activeTabIndex}
+                                selectedTabClassName={tabClassNames.tabSelected}
+                                selectedTabPanelClassName={tabClassNames.tabPanelSelected}
+                                onSelect={onActivateTab}
+
+                                // TODO: focusTabOnClick should be true for accessibility, but currently conflicts
```
kbangelov, line 391: Perhaps it could be solved by artificially making the costume the last focused in the context, when we click on something in its paint editor, because that seems to solve the problem?

kbangelov (commented):
Looks good, left a few small comments.

packages/scratch-gui/src/components/gui/gui.jsx
KManolov3: I think that comment's leftover from another PR, so probably shouldn't be handled here, given that the <Tabs> component (and this comment in particular) isn't directly related to modal behaviour.

packages/scratch-gui/src/components/library/library.jsx
```diff
@@ -402,6 +407,10 @@ class LibraryComponent extends React.Component {
                 </div>));
     }
     render () {
+        // TODO: Should the close button be focused first or last when the modal opens?
```
KManolov3: I think both of those can be considered correct behaviour, depending on the point of view. Probably leaning a bit more towards having the initial focus in "Search" for usability.

packages/scratch-gui/src/containers/library-item.jsx
```diff
@@ -48,11 +49,35 @@ class LibraryItem extends React.PureComponent {
...
+        if (this.props.disabled) {
+            return;
+        }
+
+        if (e.key === KEY.ENTER) {
             e.preventDefault();
             this.props.onSelect(this.props.id);
         }
+        /*
+        Previously, pressing SPACE would select a Costume, Sprite, Sound, or Backdrop, but there was a bug:
```
KManolov3: Nitpick: this comment adds a nice explanation, but might make sense to be removed at a later point in the a11y effort, since it references old behavior.

packages/scratch-gui/src/contexts/modal-focus-context.jsx
```diff
@@ -0,0 +1,50 @@
...
+ * @param {object} props - The component props.
+ * @param {React.ReactNode} props.children - The child components that will have access to the context.
+ * @returns {React.ReactElement} The Context Provider wrapping the children.
+ */
+export const ModalFocusProvider = ({children}) => {
+    const lastFocusedElement = useRef(null);
+
+    const captureFocus = useCallback(() => {
+        lastFocusedElement.current = document.activeElement;
+    }, []);
```
KManolov3, line 23: Nitpick: Isn't `lastFocusedElement` still technically a dependency?

KManolov3 (approved):
Clean work 🙌

packages/scratch-gui/src/containers/library-item.jsx
adzhindzhi: Yeah, I mostly included it for the reviewers in case anyone has objections to the current behavior. If not, I think I’ll just leave the current behavior in the comment and remove the old behavior part.

packages/scratch-gui/src/components/gui/gui.jsx
kbangelov: my bad

packages/scratch-gui/src/contexts/modal-focus-context.jsx
adzhindzhi: Technically, yes, but `lastFocusedElement` is a ref, and changing `.current` wouldn’t trigger a re-render. And the ref object itself doesn’t change, so it shouldn’t affect anything. Or am I missing something? 😅

kbangelov (approved):

packages/scratch-gui/src/components/library/library.jsx
adzhindzhi: Note for other reviewers: After further discussion with @KManolov3, we agreed to leave the default behavior of focusing the Back button first. We can revisit this later if we decide to change it.

packages/scratch-gui/src/lib/audio/shared-audio-context.js
```diff
@@ -10,6 +10,7 @@ if (!bowser.msie) {
     const initAudioContext = () => {
         document.removeEventListener('mousedown', initAudioContext);
         document.removeEventListener('touchstart', initAudioContext);
+        document.removeEventListener('keydown', initAudioContext);
```
adzhindzhi, line 13: You can now navigate to the Sound Editor without either `mousedown` or `touchstart` firing

packages/scratch-gui/src/components/extension-button/extension-button.jsx
```diff
@@ -73,10 +73,15 @@ const ExtensionButton = props => {
     useEffect(() => {
         if (!shouldShowFaceSensingCallouts) return;
 
-        const onFirstClick = () => {
+        const onFirstInteraction = e => {
             const isExtensionButtonVisible = document.querySelector('div[class*="extension-button-container"]');
             if (!isExtensionButtonVisible) return;
 
+            if (e.type === 'keydown') {
+                // Prevent focus from jumping to the next element in the tab order after keydown finishes
```
KManolov3, line 81: 🥲

KManolov3 (approved):
The latest changes make sense to me. It's unfortunate that the logic is as custom for every feature callout, but addressing that will be a larger effort.

Issue #418, opened 2026-01-17
Remove Face Sensing notification


medlans:
I'm referring to the notification that tells "Hey there! There is a new extension!" I feel like it doesn't make sense to have the notification anymore because it's been 3+ months since the extension was added

ManoaBrunsperger:
yeah.

nimeratus:
I still get the notification if I open Scratch while signed out (in an incognito window)

redspacecat:
> I still get the notification if I open Scratch while signed out (in an incognito window)

possibly the feature isn't quite live yet.
It probably soon will be though.

nimeratus:
> > I still get the notification if I open Scratch while signed out (in an incognito window)
> 
> possibly the feature isn't quite live yet. It probably soon will be though.

But the even newer features are working: the green flag can be focused by clicking the stop sign then pressing `Shift`+`Tab`, and clicking the flag removes the focus

PR #423, opened 2026-01-27
[UEPR-480]: Remove face sensing feature callouts

adzhindzhi:
### Resolves

- [UEPR-480](https://scratchfoundation.atlassian.net/browse/UEPR-480)

### Proposed Changes

- Remove code related to face sensing feature callouts

### Reason for Changes

- The face sensing feature has been released for more than 60 days

[UEPR-480]: https://scratchfoundation.atlassian.net/browse/UEPR-480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -189,7 +189,6 @@ const GUIComponent = props => {
         onTelemetryModalOptOut,
         onUpdateProjectThumbnail,
         showComingSoon,
-        showNewFeatureCallouts,
```
adzhindzhi, line 192: The `showNewFeatureCallouts` prop is no longer used, but I kept it in the props for future feature callouts, should we remove it as well (and also stop passing it from `www`)?

KManolov3:
This will address https://github.com/scratchfoundation/scratch-editor/issues/418

packages/scratch-gui/src/components/extension-button/extension-button.raw.css
```diff
@@ -10,41 +13,3 @@
 .driver-active .driver-overlay {
```
KManolov3: The two styles below also fit the "generic-driverjs-bandaid" mold, I think

KManolov3: Hmm.. I see that there are two files with similar styles, which will now become unused. That seems like an overkill. I propose removing them entirely with this PR, and adding an example `driverjs-example.raw.css` file (either as part of this PR, or maybe in the future) which contains the necessary styles to make a callout not block screen clicks

KManolov3 (approved):

packages/scratch-gui/src/components/extension-button/extension-button.raw.css
adzhindzhi: Agreed. To be honest, I think I’ll remove them for now. They’ll still exist in the git history, so if we ever need them again, we can bring them back. We can also discuss, as part of another PR, whether it makes sense to add an example` raw.css` file. It would definitely make sense to document this somewhere, even if we don’t include the raw files themselves, for example by providing guidance on how to work with Driver.js styles (e.g. using `.raw.css` files).

packages/scratch-gui/src/components/gui/gui.jsx
KManolov3: I'm fine with both, I propose adding a doc style comment to it justifying its existence if we decide to keep it.

PR #424, opened 2026-01-27
[UEPR-462] Ensure "Stage" and "Target" editor regions are navigable through tab navigation

kbangelov:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-462

### Proposed Changes

Adds some better keyboard accessibility
- Improved **aria-labels** and such for most focusable components
- Restricts user from focusing on other elements outside the modal when in full screen mode via **modal-focus-context.jsx**
- Makes the **popover** on **direction** input disappear on unfocus via custom hook **useFocusTrap.js**

### Does not include

- **Popover** navigation with arrows for **direction** input
- "**Choose a Sprite/Backdrop**" menu navigation with keyboard.

### Reason for Changes


packages/scratch-gui/src/hooks/useFocusOutside.js
```diff
@@ -0,0 +1,41 @@
...
+ * their interactive area (e.g. via keyboard navigation).
+ * @param {() => void} onClose
+ * Closing popover function when focus moves outside both the container and popover.
+ * @returns {object}
+ * - containerRef: reference to the container of the element that activates the popover,
+ * - popoverRef: reference to be attached to popover
+ * }}
+ */
+export default function useFocusOutside (onClose) {
+    const containerRef = useRef(null);
```
kbangelov: I am actually leaning more towards replacing this with **inputRef** or more generically **openingElementRef** that we apply to the **input** or whatever element activates the popover instead of the container including them both.

redspacecat:
keep in mind that it seems some of the recent changes relating to navigation are [causing issues](https://scratch.mit.edu/discuss/topic/863862/)

packages/scratch-gui/src/contexts/modal-focus-context.jsx
```diff
@@ -3,11 +3,15 @@ import PropTypes from 'prop-types';
...
 export const ModalFocusContext = createContext(null);
 
+const ALLOW_SELECTOR = '[data-focusable]';
+
 /**
  * A context provider that manages focus restoration strategies for modals.
  *
  * It keeps track of the element that was focused prior to a modal opening (`captureFocus`)
  * and attempts to restore focus to that element when the modal closes (`restoreFocus`).
+ * It can make all other elements outside the modal unfocusable via tab (`restrictFocusableElements`)
```
KManolov3: This explanation might be better served in the jsdoc by describing the `value` object member properties

packages/scratch-gui/src/contexts/modal-focus-context.jsx
```diff
@@ -29,12 +33,53 @@ export const ModalFocusProvider = ({children}) => {
...
+
+    const restrictFocusableElements = useCallback(() => {
+        const allElements = document.body.querySelectorAll('*');
+
+        allElements.forEach(el => {
+            if (!el.matches(ALLOW_SELECTOR)) {
+                makeUnfocusable(el);
+            }
+        });
+    }, []);
```
adzhindzhi: Hmm, I like the idea of trapping focus inside the stage when it's in full screen, but making every single element have `tabIndex=-1` feels a bit hacky to me...  I've been thinking about other ways to achieve the same behavior (I am basically thinking out loud here, so sorry if my thoughts are a bit messy). 

The main issue is that this isn't a real modal, but we want to treat it like one. I don't see using an actual dialog as a viable option though. But maybe instead of manually changing the `tabIndex`, we could make some sort of a `trapFocus`/`restrictFocus` helper that targets a specific container, listens to keydown events, and:
- Collects a list of focusable items within the container (we could potentially reuse your existing `data-focusable` prop)
- If focus is outside that list, move it back to the first focusable element
- If the last focusable element is focused and `Tab` is pressed, loop focus back to the first
- If the first element is focused and `Shift + Tab` is pressed, move focus to the last

I'm fine with using `ModalFocusProvider` for now and handling this logic here, but generally it might make sense to extract it on its own at some point, as I can see this potentially being used in other places as well, which are not necessarily modals. The benefits I see in this approach are that if you have multiple containers or areas which can trap focus, you can "specify" the target region, in which the focus is trapped. Right now, if any other element outside the Stage has `data-focusable` as a prop, we'd still be able to navigate it even when in `fullScreen`.

What do you think? I'd like to here @KManolov3's thoughts as well. There's something similar (although not entirely) implemented [here](https://css-tricks.com/a-primer-on-focus-trapping/), which I think could be helpful.

packages/scratch-gui/src/components/green-flag/green-flag.jsx
```diff
@@ -1,33 +1,35 @@
...
-
+    const intl = useIntl();
     return (
         <button
             className={styles.greenFlagButton}
-            onClick={handleOnClick}
+            onClick={onClick}
             {...componentProps}
+            aria-label={intl.formatMessage(startProjectMessage)}
+            {...(isFullScreen ? {'data-focusable': true} : {})}
```
adzhindzhi: Definitely just a nitpick, and totally fine to leave as-is, but usually the destructured `...componentProps` go last 😅 That way we can override already-passed props if needed, for example `aria-label`.

packages/scratch-gui/src/components/green-flag/green-flag.jsx
```diff
@@ -1,33 +1,35 @@
...
+import {defineMessage, useIntl} from 'react-intl';
 
 import greenFlagIcon from './icon--green-flag.svg';
 import styles from './green-flag.css';
 
+const startProjectMessage = defineMessage({
+    id: 'gui.aria.startProjectButton',
+    defaultMessage: 'Start project',
+    description: 'accessibility label for start project button'
+});
```
adzhindzhi, line 13: Maybe we can run this by Tereza or someone just to make sure that we'd want to use `Start project` for the accessibility label? As opposed to `Green Flag` for example?

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -73,6 +74,23 @@ const StageHeaderComponent = function (props) {
...
+        unrestrictFocusableElements
+    } = useContext(ModalFocusContext);
+
+    useLayoutEffect(() => {
+        if (isFullScreen) {
+            captureFocus();
+            restrictFocusableElements();
+        } else {
+            unrestrictFocusableElements();
+            restoreFocus();
```
adzhindzhi: Do we actually need to `capture`/`restore` the focus? The focused button should be visible both when you are in full screen, and when you are not, so shouldn't the focus be preserved anyway?

packages/scratch-gui/src/components/stop-all/stop-all.jsx
```diff
@@ -1,23 +1,36 @@
...
+
+    const intl = useIntl();
     return (
         <button
             className={styles.stopAllButton}
             onClick={onClick}
             {...componentProps}
+            aria-label={intl.formatMessage(stopProjectMessage)}
+            aria-disabled={!active}
+            {...(isFullScreen ? {'data-focusable': true} : {})}
```
adzhindzhi: Same nitpick as for the green flag button

packages/scratch-gui/src/components/stop-all/stop-all.jsx
```diff
@@ -1,23 +1,36 @@
...
     } = props;
+
+    const intl = useIntl();
     return (
         <button
             className={styles.stopAllButton}
             onClick={onClick}
             {...componentProps}
+            aria-label={intl.formatMessage(stopProjectMessage)}
+            aria-disabled={!active}
```
adzhindzhi: We could probably just use `disabled` and actually disable the button, instead of just changing the styles and adding `aria-disabled`? We probably should have done this when we initially switched to a button anyway 😅

packages/scratch-gui/src/hooks/useFocusOutside.js
```diff
@@ -0,0 +1,41 @@
+import {useEffect, useRef} from 'react';
```
adzhindzhi: To be honest, this feels a bit like an overkill to me, and I'm not a huge fan of introducing very custom and highly specific logic/hooks for this... I am worried it could become harder to maintain over time. Maybe a simpler approach would work, for example handling keydown events directly in the `direction-picker` component and closing the modal when `Tab` or `Shift + Tab` is pressed? Again, I'd like to hear @KManolov3's thoughts as well, but I think keeping it as simple as possible would be the better choice, at least for this.

packages/scratch-gui/src/components/green-flag/green-flag.jsx
kbangelov: sure

kbangelov: She agreed that since the blockly element "When _green flag img_ clicked" is not readable for the moment, it makes more sense to go with "Start project"

packages/scratch-gui/src/components/stage-header/stage-header.jsx
kbangelov: Not sure I understand the question, this is part of the limiting focus logic, although the comment will probably be resolved once I answer the one above.

packages/scratch-gui/src/components/stop-all/stop-all.jsx
kbangelov: Yeah, very fair.

packages/scratch-gui/src/hooks/useFocusOutside.js
kbangelov: perhaps, it was a little bit 50/50 for me. I'll also wait to hear @KManolov3's opinion.

packages/scratch-gui/src/contexts/modal-focus-context.jsx
kbangelov: I agree about the flaw of having other elements with the same prop. I missed that. Besides that my code seems to match most of your ideas, except the way we make them unfocusable by tab via **tabIndex=-1** which does feel just a little bit hacky. I agree with that too.

My suggestion for the first issue is to add a param to **data-focusable** so different sections don't get mixed up and call the context with that param. Making this into a hook, instead of context should be considered, then using the container of the items to look for all the direct focusables that should remain. I'm even leaning more in that direction right now.

The **tabIndex** logic itself doesn't feel terribly wrong to me since that's the prop is responsible for the focus of elements. I am struggling to picture something different right now. But I'll also wait to hear @KManolov3 on this one.

KManolov3: A general thought I'm having with accessibility related logic is to try to keep it as simple as possible due to both maintenance and testing concerns.

With that in mind, I think a `trapFocus` approach, where you have a hook to which you pass a given parent element and a data prop name makes sense to me. The hook can return a `trapFocus()` and `releaseFocus()` methods. `trapFocus()` would then add event listeners which listen for Tab events and iterate over the children of the parent element which have the specified data prop name.

That should be fairly concise in terms of implementation and it'd allow us to avoid setting a prop for all elements.

What do you think?

packages/scratch-gui/src/hooks/useFocusOutside.js
KManolov3: The hook implementation looks fairly clean, but I think I agree with what Ayshe has proposed in terms both of brevity and of keeping the behavior closer to where it's being used. It'd also allow us to not depend on refs. Let's collect more similar cases to better understand whether it'd be something we'd want to generalize (and how it'd be best to do so), before considering extracting a hook.

packages/scratch-gui/src/contexts/modal-focus-context.jsx
kbangelov: I'll give it a shot.

packages/scratch-gui/src/components/direction-picker/direction-picker.jsx
```diff
@@ -51,15 +51,40 @@ const messages = defineMessages({
...
+
+            props.onClosePopover();
+        };
+
+        document.addEventListener('focusin', handleFocusIn);
+
+        return () => {
+            document.removeEventListener('focusin', handleFocusIn);
+        };
+    }, [containerRef, props.onClosePopover]);
```
adzhindzhi: I think we can remove `containerRef` from the props, since it doesn't change.

packages/scratch-gui/src/components/green-flag/green-flag.jsx
```diff
@@ -1,32 +1,34 @@
...
-    }, [onClick]);
-
+    const intl = useIntl();
     return (
         <button
             className={styles.greenFlagButton}
-            onClick={handleOnClick}
+            onClick={onClick}
+            aria-label={intl.formatMessage(startProjectMessage)}
+            {...(isFullScreen ? {'data-focusable': true} : {})}
```
adzhindzhi: nitpick: Isn't this the same as `data-focusable={isFullScreen || undefined}`? Same for the stop button?

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -73,6 +75,19 @@ const StageHeaderComponent = function (props) {
...
+
+    useLayoutEffect(() => {
+        if (isFullScreen) {
+            trapFocus();
+        } else {
+            releaseFocus();
+        }
+
+        return releaseFocus;
+    }, [isFullScreen, trapFocus, releaseFocus]);
```
adzhindzhi: Instead of doing this in a `useEffect`, can we call `trapFocus`/`releaseFocus` when the full screen button gets clicked? I think it's a bit more explicit, and we don't risk triggering it more times than we actually need to?

packages/scratch-gui/src/hooks/useFocusTrap.js
```diff
@@ -0,0 +1,69 @@
+import {useCallback} from 'react';
+import {KEY} from '../lib/navigation-keys';
+
+/**
+ * Hook to trap focus within a container element.
+ * @param {{current: HTMLElement}} containerRef - Parent container ref
+ * @param {string} dataAttribute - Data attribute used to mark focusable elements (default: 'data-focusable')
+ * @returns {{trapFocus: () => void, releaseFocus: () => void}} - Functions to activate and deactivate the focus trap
+ */
+const useFocusTrap = (containerRef, dataAttribute = 'data-focusable') => {
```
adzhindzhi: Let's make the `dataAttribute` required and pass it explicitly. Otherwise, we somewhat have two levels of abstraction, and if we need to use `useFocusTrap` again, we might missing adding the `data-focusable` attribute where necessary.

packages/scratch-gui/src/hooks/useFocusTrap.js
```diff
@@ -0,0 +1,69 @@
...
+    const getFocusableElements = useCallback(() => {
+        const container = containerRef.current;
+        if (!container) return [];
+
+        return Array.from(
+            container.querySelectorAll(`[${dataAttribute}]`)
+        ).filter(el => !el.hasAttribute('disabled'));
```
adzhindzhi: You can use `el.disabled` directly.

packages/scratch-gui/src/hooks/useFocusTrap.js
```diff
@@ -0,0 +1,69 @@
...
+    const trapFocus = useCallback(() => {
+        const container = containerRef.current;
+        if (!container) return;
+
+        document.addEventListener('keydown', handleKeyDown);
+
+        const focusable = getFocusableElements();
+        if (focusable.length) {
+            focusable[0].focus();
+        }
```
adzhindzhi: I think this part is unnecessary. Let the first `Tab` handle the focus. Currently, we focus the "Green flag" button, but no focus indicator is shown, so it's a bit weird. I think it makes more sense to keep the default behavior - not focus anything manually when we change to full screen, then pressing `Tab` focuses the first element, which is the Green Flag button

packages/scratch-gui/src/components/stage-header/stage-header.jsx
kbangelov: I mean clicking button changes the state and activates this just once, I think. I think it's safe to keep it this way. It seems more tied to the state rather than a click.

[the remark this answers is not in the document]
kbangelov: not if we add the attribute

adzhindzhi: Well, not exactly. For example, when you are in embed mode, `scratch-www` renders the editor with `isFullScreen=true`, which will trap the focus, and never release it. However, we wouldn't want to trap the focus when we're in embed mode, because we'd prevent any other elements in the document from being navigable.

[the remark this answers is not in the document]
adzhindzhi: But we haven't added the attribute? I think we currently only show branding in `embed` mode. With the current implementation, we can't focus it. That would get resolved if we address the other outstanding comment for trapping the focus on button click (because then we won't trap focus in embed mode at all), but I think it'd still make sense to add the attribute to the branding as well, since it technically is a focusable element.

packages/scratch-gui/src/components/green-flag/green-flag.jsx
```diff
@@ -27,6 +37,8 @@ const GreenFlagComponent = function (props) {
         <button
             className={styles.greenFlagButton}
             onClick={handleOnClick}
+            aria-label={intl.formatMessage(startProjectMessage)}
+            data-focusable={isFullScreen || null}
```
adzhindzhi, line 41: nitpick: Let's use `undefined` when we want to say that the value is not provided. It's technically the same, but `undefined` makes more sense semantically.

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -124,12 +145,18 @@ const StageHeaderComponent = function (props) {
             </div>
         );
         header = (
-            <Box className={styles.stageHeaderWrapperOverlay}>
+            <Box
+                className={styles.stageHeaderWrapperOverlay}
+                componentRef={componentRef}
```
adzhindzhi: nitpick: I think you can pass containerRef directly, React will set it properly out of the box, you don't have to do it manually

packages/scratch-gui/src/hooks/useFocusTrap.js
```diff
@@ -0,0 +1,67 @@
+import {useCallback} from 'react';
+import {KEY} from '../lib/navigation-keys';
+
+/**
+ * Hook to trap focus within a container element.
+ * @param {{current: HTMLElement}} containerRef - Parent container ref
+ * @param {string} dataAttribute - Data attribute used to mark focusable elements (default: 'data-focusable')
```
adzhindzhi: Let's update the docs, since we no longer have a default value

packages/scratch-gui/src/hooks/useFocusTrap.js
```diff
@@ -0,0 +1,67 @@
...
+ * @returns {{trapFocus: () => void, releaseFocus: () => void}} - Functions to activate and deactivate the focus trap
+ */
+const useFocusTrap = (containerRef, dataAttribute) => {
+    const getFocusableElements = useCallback(() => {
+        const container = containerRef.current;
+        if (!container) return [];
+
+        return Array.from(
+            container.querySelectorAll(`[${dataAttribute}]`)
+        ).filter(el => !el.disabled && el.offsetParent !== null);
```
adzhindzhi: When will `offsetParent` be `null`?

packages/scratch-gui/src/hooks/useFocusTrap.js
```diff
@@ -0,0 +1,67 @@
...
+            last.focus();
+        }
+    }, [getFocusableElements, containerRef]);
+
+    const trapFocus = useCallback(() => {
+        const container = containerRef.current;
+        if (!container) return;
+
+        document.addEventListener('keydown', handleKeyDown);
+    }, [handleKeyDown, getFocusableElements]);
```
adzhindzhi: Do we need `getFocusableElements` in the `deps` array?

packages/scratch-gui/src/components/green-flag/green-flag.jsx
kbangelov: linter doesn't fancy undefined

packages/scratch-gui/src/hooks/useFocusTrap.js
kbangelov: It's more of an edge case but if an element is invisible due to stylings, such as **display : none**, it should be filtered out by this check. I wonder whether to leave it, to be honest.

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -1,6 +1,6 @@
 import {FormattedMessage, defineMessages, useIntl} from 'react-intl';
 import PropTypes from 'prop-types';
-import React, {useCallback} from 'react';
+import React, {useCallback, useRef, useLayoutEffect} from 'react';
```
adzhindzhi: `useLayoutEffect` is no longer used, right?

packages/scratch-gui/src/components/direction-picker/direction-picker.jsx
```diff
@@ -51,15 +51,42 @@ const messages = defineMessages({
 
 const DirectionPicker = props => {
     const intl = useIntl();
+
+    const containerRef = useRef(null);
+    const popoverRef = useRef(null);
+
+    useEffect(() => {
+        if (!props.popoverOpen) return;
```
adzhindzhi, line 60: `props.popoverOpen` is missing from the dependency array, which will probably cause this not to work if `popoverOpen` is `false` on initial render.

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -73,6 +75,19 @@ const StageHeaderComponent = function (props) {
     } = props;
     const intl = useIntl();
 
+    const containerRef = useRef(null);
+    const {trapFocus, releaseFocus} = useFocusTrap(containerRef, 'data-focusable');
+
+    const handleEnterFullScreen = useCallback(() => {
+        onSetStageFull();
+        trapFocus();
+    }, [onSetStageFull, trapFocus]);
```
adzhindzhi, line 86: Now that we've switched to trapping the focus when the button is pressed `containerRef` will always be `null` when this executes, since container is rendered conditionally when `isFullScreen` is `true`, and this runs before `isFullScreen` becomes `true`.

packages/scratch-gui/src/hooks/useFocusTrap.js
adzhindzhi: If an element is invisible, it probably shouldn't have the `data-focusable` attribute. As far as I can see, `offsetParent` can also be `null` when the element has `position: fixed` (see [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent)).

I think it's reasonable to leave handling of hidden or fixed elements to the consumer of the hook. They may want those elements to remain focusable, or they may not. That doesn't feel like a general concern this hook needs to address... 🤔

adzhindzhi (approved):
Looks great! 🙌  I really like how this turned out!

packages/scratch-gui/src/components/direction-picker/direction-picker.jsx
```diff
@@ -51,15 +51,42 @@ const messages = defineMessages({
 
 const DirectionPicker = props => {
     const intl = useIntl();
+
+    const containerRef = useRef(null);
+    const popoverRef = useRef(null);
+
+    useEffect(() => {
```
KManolov3, line 59: Nitpick: We can leave a comment like
`// Close the popover once we focus out of it`
for readibility

packages/scratch-gui/src/hooks/use-focus-trap.js
```diff
@@ -0,0 +1,67 @@
+import {useCallback} from 'react';
+import {KEY} from '../lib/navigation-keys';
+
+/**
+ * Hook to trap focus within a container element.
+ * @param {{current: HTMLElement}} containerRef - Parent container ref
+ * @param {string} dataAttribute - Data attribute used to mark focusable elements
+ * @returns {{trapFocus: () => void, releaseFocus: () => void}} - Functions to activate and deactivate the focus trap
+ */
+const useFocusTrap = (containerRef, dataAttribute) => {
```
KManolov3, line 10: `useFocusTrap.js` is camelCase, while the other existing hook - `use-menu-navigation` is snake case. Let's choose a single naming convention for hook and stick to it. Looking at other files in the editor, snake case might be more appropriate.

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -110,8 +128,10 @@ const StageHeaderComponent = function (props) {
             <div className={styles.unselectWrapper}>
                 <Button
                     className={styles.stageButton}
-                    onClick={onSetStageUnFull}
+                    onClick={handleExitFullScreen}
                     onKeyPress={onKeyPress}
+                    aria-label={intl.formatMessage(messages.unFullStageSizeMessage)}
```
KManolov3: I think having this message be added as both aria-label and `alt` attribute might result in a double announcement from the screen reader. Let's leave it in one place only. I'm fine with the idea of having it on the button, but, as this PR is more tab navigation focused, the least-resistance change might be to leave it as it was.

packages/scratch-gui/src/hooks/use-focus-trap.js
```diff
@@ -0,0 +1,67 @@
...
+    const releaseFocus = useCallback(() => {
+        document.removeEventListener('keydown', handleKeyDown);
+    }, [handleKeyDown]);
+
+    return {
+        trapFocus,
+        releaseFocus
+    };
+};
+
```
KManolov3, line 66: Very elegant!

KManolov3 (approved):
Left a few minor comments, but otherwise looks good to me!

packages/scratch-gui/src/hooks/use-focus-trap.js
kbangelov: yes mb

KManolov3 (approved):

PR #430, opened 2026-02-02
[UEPR-501] Green flag button receives focus and prevents the project from capturing keyboard events

kbangelov:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-501

### Proposed Changes

Unfocuses green flag on click, so project stage can capture keyboard events.

### Reason for Changes

Previously it prevented project stage to capture keyboard events.

### DIFF
--- a/packages/scratch-gui/src/components/green-flag/green-flag.jsx
+++ b/packages/scratch-gui/src/components/green-flag/green-flag.jsx
@@ -1,6 +1,6 @@
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, {useCallback} from 'react';
 
 import greenFlagIcon from './icon--green-flag.svg';
 import styles from './green-flag.css';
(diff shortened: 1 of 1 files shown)


packages/scratch-gui/src/components/green-flag/green-flag.jsx
```diff
@@ -13,10 +13,20 @@ const GreenFlagComponent = function (props) {
...
     } = props;
+
+    // Unfocus so project stage can capture keyboard events
+    const handleOnClick = useCallback(e => {
+        onClick(e);
+
+        const target = e.currentTarget;
+        if (target) target.blur();
+    }, [onClick]);
+
```
KManolov3, line 25: Nitpick: double newline

packages/scratch-gui/src/components/green-flag/green-flag.jsx
```diff
@@ -13,10 +13,20 @@ const GreenFlagComponent = function (props) {
         title,
         ...componentProps
     } = props;
+
+    // Unfocus so project stage can capture keyboard events
```
KManolov3: Nitpick: can we add slightly more clarification why this is important (e.g. "so blocks such as "when X key is pressed" are able to be triggered when project is played from "Green Flag" button")

KManolov3 (approved):

PR #433, opened 2026-02-03
[UEPR-483] "Choose a Sprite" and "Choose a Backdrop" keyboard accessibility

kbangelov:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-483

### Proposed Changes

- Adds keyboard navigation logic.
- **UI bug**: tooltips don't refresh in case of first open or **shift+tab**-ing back to the menu.

### Reason for Changes


packages/scratch-gui/src/hooks/use-action-menu-navigation.js
```diff
@@ -0,0 +1,153 @@
...
+ * the menu receives focus. If `null`, the last menu item will be focused.
+ * @returns {object} Navigation helpers and refs for use in an action menu component.
+ * - containerRef - Ref to the container element that holds all menu items.
+ * - buttonRef - Ref to the main action button.
+ * - isExpanded - Current expanded state of the menu.
+ * - setIsExpanded - Setter to manually expand/collapse the menu.
+ * - handleKeyDown - Keydown handler to navigate through menu items with arrow keys and handle tab.
+ * - handleOnFocus - Focus handler to expand the menu and focus the default item.
+ */
+export default function useActionMenuNavigation (
```
adzhindzhi: I'm not a fan of introducing another custom hook for such a specific use case. In my opinion, extracting logic into a hook makes sense when there's clear potential for reuse. I don't see this ever being reused, since it's tightly coupled to the `ActionMenu` component. I think we could approach this problem from a simpler perspective. As I understand it, we're trying to do two things:

- Open/close the action menu on focus/blur

    We already have open/close logic implemented for mouse enter/leave. We could easily reuse that behavior by adding:
    ```
    onFocus={this.handleToggleOpenState}
    onBlur={this.handleClosePopover}
    ```
    to the container holding the buttons and tooltips.
    
- Enable `ArrowUp`/`ArrowDown` navigation

    I believe this can be handled in a much simpler way as well. Right now we're relying on `data-*` quite a lot (both in this, and other PRs I am seeing) and introducing quite a bit of custom logic, which may become harder to maintain over time. Instead, we could handle `ArrowUp`/`ArrowDown` in an `onKeyDown` handler on the container and use button refs directly. In the navbar PR, we were trying to avoid that, because it generated a lot of boilerplate.. I don't see that being the case here. 

Both of these would remove the need for a separate BFS traversal and manual handling of `Tab`/`Shift+Tab`, and would keep the implementation much cleaner. If, at some point in the future, we need this arrow navigation behavior elsewhere, that would be a good time to extract it into something like `useArrowNavigation` hook. For now, though, I don't see a strong reason to introduce that level of abstraction. Overall, for accessibility-related work, I think we should aim to keep the implementation as simple as possible and avoid introducing unnecessary custom logic, since it increases the risk of regressions (unless there's a clear need for it, as in the navbar PR, for example).

kbangelov: I agree, I have refactored it back inside. Only thing is that both the **onFocus** and **onBlur** have very specific behaviours, such as focusing on the last element immediately on focusing the menu as well as the blur on **shift+tab** custom logic. So I have not added **onFocus** and **onBlur** in the component's wrapper element.

packages/scratch-gui/test/integration/sounds.test.js
```diff
@@ -173,7 +173,8 @@ describe('Working with sounds', () => {
 
         await loadUri(uri);
         await clickText('Sounds');
-        const el = await findByXpath('//button[@aria-label="Choose a Sound"]');
+        // Just select any button inside the tab so the controls impact the sounds editor
+        const el = await findByXpath('//button[@aria-label="Delete"]');
```
adzhindzhi: Why did we need to change this?

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,162 +1,249 @@
...
+    const [forceHide, setForceHide] = useState(false);
+
+    const closeTimeoutRef = useRef(null);
+    const mainTooltipId = useRef(`tooltip-${Math.random()}`).current;
+
+    const containerRef = useRef(null);
+    const buttonRef = useRef(null);
+    const [isExpanded, setIsExpanded] = useState(false);
+    const skipNextBlurRef = useRef(false);
+    const itemRefs = useRef([]);
```
adzhindzhi, line 26: nitpick: Can we order these a bit, e.g. first the `states`, then the `refs`? It's a bit easier to track if anything got lost during the refactoring.

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,162 +1,238 @@
...
+    const handleOnFocus = useCallback(() => {
+        setIsExpanded(true);
+        const items = itemRefs.current;
+        if (!items.length) return;
+
+        // default to last item (first above)
+        const defaultItem = items[items.length - 1];
+        focusItem(defaultItem);
+        // TODO: refresh tooltip so it repositions correctly
+    }, [itemRefs, focusItem, setIsExpanded]);
```
adzhindzhi: Do we actually need this + the `useEffect` on lines 28-32? Let the main button be focused, it has a default action anyway so it seems correct to be able to select the same that you would by simply clicking on the main button. Then, if you press ArrowUp/Down keys, you'll enter the menu and start navigating through the items.

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,162 +1,238 @@
...
+                    const tabindex = el.getAttribute('tabindex');
+                    if (tabindex === '-1') return false;
+                    return true;
+                });
+                const currentIndex = filteredFocusables.indexOf(buttonRef.current);
+                if (currentIndex > 0) {
+                    filteredFocusables[currentIndex - 1].focus();
+                }
+            }
+            return;
```
adzhindzhi: This logic seems redundant as well, onFocus/onBlur should already handle `Tab`/`Shift + Tab` presses

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,162 +1,238 @@
...
+                        ) => {
                             const isComingSoon = !handleClick;
                             const hasFileInput = fileInput;
-                            const tooltipId = `${this.mainTooltipId}-${title}`;
+                            const tooltipId = `${mainTooltipId}-${title}`;
                             return (
-                                <li key={`${tooltipId}-${keyId}`}>
+                                <li
+                                    key={`${tooltipId}-${keyId}`}
+                                    tabIndex={-1}
```
adzhindzhi: Do you need the `tabIndex={-1}` here? Is the `li` focusable by default?

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,162 +1,238 @@
...
+    const handleKeyDown = useCallback(e => {
+        switch (e.key) {
+        case KEY.ARROW_DOWN:
+            e.preventDefault();
+            handleMove(1);
+            break;
+        case KEY.ARROW_UP:
+            e.preventDefault();
+            handleMove(-1);
+            break;
```
adzhindzhi: nitpick: This can probably be simplified if we do something like:
```
if (e.key === KEY.ARROW_DOWN || KEY.ARROW_UP) {
  e.preventDefault();

  const direction = KEY.ARROW_UP ? 1 : -1;
  handleMove(direction);
}
```

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,162 +1,238 @@
...
+    
+    useEffect(() => {
+        const buttonEl = buttonRef.current;
+        if (!buttonEl) return;
+
+        buttonEl.addEventListener('touchstart', handleTouchStart);
+        return () => {
+            buttonEl.removeEventListener('touchstart', handleTouchStart);
+        };
+    }, [handleTouchStart]);
```
adzhindzhi: Can we move this on top right after the other `useEffect`, it's easier to track them that way. I even think we could merge the two into one `useEffect` that runs when the component mounts (as otherwise we'd attach multiple event listeners), but I'm not entirely sure about that.

packages/scratch-gui/test/integration/sounds.test.js
kbangelov: It is quite mysterious to me as well why the test failed when the **Choose a Sound** action menu was selected for the test. The logic of the test was "Just click any object so the controls impact the sound editor canvas".

packages/scratch-gui/src/components/action-menu/action-menu.jsx
kbangelov: It may appear that way but I tried a few different things and here's where the issue comes from.

In case we want to **shift+tab** from an inside element, the focus goes to the root and back to the first element respectfully, instead of the previous element when comparing to a menu. It does that because that's the previous element that appears in the structure. If we add a handler for the entire wrapper, it will recognize the inside **tab** but still execute the forementioned sequence. If we prevent default, it will prevent default generally, so this is the only way that I found to work. Ideas are welcome

packages/scratch-gui/src/components/action-menu/action-menu.jsx
kbangelov: No, I guess not.

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,152 +1,193 @@
...
+
+    useEffect(() => {
+        if (!isExpanded) {
+            buttonRef?.current?.blur();
+        }
+    }, [isExpanded, buttonRef]);
```
adzhindzhi: Do we still need this?

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,152 +1,193 @@
...
+            onFocus={handleToggleOpenState}
+            onBlur={handleClosePopover}
+            ref={containerRef}
+        >
+            <button
+                aria-label={mainTitle}
+                className={classNames(styles.button, styles.mainButton)}
+                data-for={mainTooltipId}
+                data-tip={mainTitle}
+                onFocus={handleOnFocus}
```
adzhindzhi: Do we need the `onFocus` for the `button` as well? It should already be handled by the `onFocus` in its parent component?

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,152 +1,193 @@
...
+
+    const handleKeyDown = useCallback(e => {
+        if (e.key === KEY.ARROW_DOWN || e.key === KEY.ARROW_UP) {
+            e.preventDefault();
+            const direction = e.key === KEY.ARROW_UP ? -1 : 1;
+            handleMove(direction);
+        } else if (e.key === KEY.TAB) {
+            setIsExpanded(false);
+            focusItem(buttonRef.current);
         }
```
adzhindzhi, line 136: I don't think we need to handle tab presses manually here, it should do this out of the box. The container will get blurred, which will cause the menu will close anyway.

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,152 +1,193 @@
...
-    }
-    shouldComponentUpdate (newProps, newState) {
-        // This check prevents re-rendering while the project is updating.
-        // @todo check only the state and the title because it is enough to know
-        //  if anything substantial has changed
-        // This is needed because of the sloppy way the props are passed as a new object,
-        //  which should be refactored.
-        return newState.isOpen !== this.state.isOpen ||
-            newState.forceHide !== this.state.forceHide ||
-            newProps.title !== this.props.title;
```
adzhindzhi, line 43: Is this still a valid concern now that we've switched to a function component? I think it is, and we'll cause multiple re-renders with the current implementation, since we no longer check this. We should either:
- Wrap the component in a `React.memo` (which I think is the alternative of `shouldComponentUpdate` for function components) and pass a callback function, which only checks if the title has changed. The state change will trigger a re-render in any case. We can then bring back the initial comment and original TODO there.
- Change the way the props are passed, which is what the original TODO suggests

packages/scratch-gui/test/integration/sounds.test.js
adzhindzhi: Hmmm, that's weird. Nothing in the current implementation should affect this. Are they still failing? Maybe it was a temporary thing with the previous changes?

packages/scratch-gui/src/components/action-menu/action-menu.jsx
kbangelov: We don't need it anymore, yes.

packages/scratch-gui/src/components/action-menu/action-menu.jsx
kbangelov: Actually we do need to focus the core item in order to work for **Shift + tab**, otherwise the previous element will be that core button.

packages/scratch-gui/test/integration/sounds.test.js
kbangelov: Brought it back. Seems to work.

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -184,29 +218,34 @@ class ActionMenu extends React.Component {
...
     onClick: PropTypes.func.isRequired,
     title: PropTypes.node.isRequired,
     tooltipPlace: PropTypes.string
 };
 
-export default ActionMenu;
+export default React.memo(ActionMenu, (prevProps, nextProps) =>
+    // Only re-render if the title changes
+    prevProps.title === nextProps.title
+);
```
adzhindzhi, line 290: Can we also bring back the comment on why this was needed in the first place? Thanks!

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,152 +1,182 @@
...
+    const handleKeyDown = useCallback(e => {
+        if (e.key === KEY.ARROW_DOWN || e.key === KEY.ARROW_UP) {
+            e.preventDefault();
+            const direction = e.key === KEY.ARROW_UP ? -1 : 1;
+            handleMove(direction);
+        } else if (e.key === KEY.TAB || e.key === KEY.ESCAPE) {
+            setIsExpanded(false);
+            focusItem(buttonRef.current);
+        }
+    }, [handleMove, isExpanded, setIsExpanded]);
```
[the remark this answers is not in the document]
adzhindzhi: I'd say adding a few test cases to cover this makes sense, but we don't necessarily need to do it in this PR.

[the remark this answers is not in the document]
kbangelov: Will do in a separate PR

adzhindzhi (approved):
Looks good to me! 👍

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,152 +1,185 @@
...
         // Mouse enter back in after timeout was started prevents it from closing.
-        if (this.closeTimeoutId) {
-            clearTimeout(this.closeTimeoutId);
-            this.closeTimeoutId = null;
-        } else if (!this.state.isOpen) {
-            this.setState({
-                isOpen: true,
-                forceHide: false
-            });
+
```
KManolov3: Nitpick: there's now an empty line between the comment and the code it's referring to

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,152 +1,185 @@
...
+    }, [handleMove, isExpanded, setIsExpanded]);
+
+    const handleClosePopover = useCallback(() => {
+        closeTimeoutRef.current = setTimeout(() => {
+            setIsExpanded(false);
+            closeTimeoutRef.current = null;
+        }, CLOSE_DELAY);
+    }, []);
+
+    const clickDelayer = useCallback(
```
KManolov3: This looks like it slightly differs in behavior from production:
- on production, when we open and close the sprite modal, the action menu is closed
- with this change, when we open and close the sprite modal, the action menu is still open

I am not averse to the updated behavior, as it aligns with the "return focus to previously selected element when you navigate out of a modal". 

However there are some visually unpleasant cases:
- if I open the "Upload a sprite" menu from the action button, the menu closes after a timeout, but if I navigate out of the upload dialog which it opens, it reopens the menu, but the label shows on the bottom
- same for other actions

I am open to suggestions for what the most consistent behavior (and the least effort change) would be.

As a fallback, keeping the production behavior (and possibly making sure the inner menu items are also unselected, so that no inconsistent label appears) can be an option for this case (so, blur out of the focused action menu, but keep the focus, so that TAB would return to the same / next item).

Image

KManolov3: Additionally, the behavior when clicking "Surprise" has changed here - it unfocuses the action menu entirely, while on production it blurs, but keeps the focus on the menu item.

KManolov3 (commented):
Looks mostly good, left a comment on an edge case. Thanks for migrating the action menu to React hook syntax!

packages/scratch-gui/src/components/action-menu/action-menu.jsx
kbangelov: For the first comment, I have added a timer that blurs and refocuses on the menu item so to avoid the visual unpleasantries of the positioning of the tooltip upon expansion (the issue was that it needed time to expand so when it refocused the menu was not open yet and the tooltip got stuck in the middle of the core button and the right one)

kbangelov: For the second comment I altered the behavior of the former **clickDelayer** and removed all the unfocus logic since it came from the idea of closing down the menu upon click which we don't want anymore. This, in my opinion, remains more consistent with the idea of keeping focus on the selected item after its effect has been fulfilled, that we have implemented so far in the accessibility tasks.

What do you think @KManolov3? cc: @adzhindzhi

adzhindzhi: FYI, we still have an issue related to the first case. If you select `Choose a Sprite`, then close the modal, the tooltip is displayed twice - once for the main button and once for the `Choose a Sprite` item in the menu.

Regarding the `clickDelayed` issue, I don't have a strong opinion, but I think the current production behavior, where the menu closes on click while focus remains on the main button, feels more convenient than keeping the menu open. That said, I can see benefits to both approaches.

In general, the behavior when clicking on menu items feels a bit inconsistent right now. For example:
- For `Surprise` and `Paint`, the menu stays open and the selected item remains focused.
- For `Choose a Sprite` and `Upload`, the menu closes, and when the modal is closed, the menu is re-opened and the main button is focused (and in the case of `Choose a Sprite`, two tooltips are shown).

It might be better to standardize this behavior across all menu items. For example, the menu could always close while keeping focus on the main button, or we could choose another consistent option if this one doesn't make sense or introduces technical complications. What do you think?

kbangelov:
The latest commit fixes the following 3 bugs that existed:
- Escape used to not show the tooltip for the core element
- if we selected a sprite from inside the menu, it showed a double tooltip on the way back
- clicking escape on the tooltip selector library used to also close the action menu on the way back

What I did was remove **onBlur** logic and replace it in the following way - **onBlur** unites a few of the following cases:
- clicking somewhere else (mousedown works now, unlike touchstart)
- we were also already handling **tab**
- the one blur motion we don't handle because we don't want to is after clicking on a menu option that takes us to library or file upload, where we now don't close the menu, which avoids most (hopefully all) bugs.

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,152 +1,236 @@
...
+                setIsExpanded(true);
+                // wait to expand before moving focus to menu item and displaying tooltip
+                setTimeout(() => {
+                    handleMove(direction);
+                }, CLOSE_DELAY);
+            }
+        } else if (e.key === KEY.TAB) {
+            setIsExpanded(false);
+            focusItem(buttonRef.current);
+        } else if (e.key === KEY.ESCAPE) {
```
KManolov3, line 134: I see some possibly unintended behavior when clicking `ESCAPE`, after which attempting to `TAB`- I need to TAB twice for something to happen. It's more of a corner case, but might be worth fixing if it isn't too hard. Otherwise I am fine with leaving as is for now.

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,152 +1,236 @@
...
-        return event => {
+    }, [isExpanded, focusItem, itemRefs.current.length]);
+
+    useEffect(() => {
+        if (!isExpanded) {
             ReactTooltip.hide();
+        }
+    }, [isExpanded]);
+
+    const clickEvent = useCallback(
```
KManolov3: Does this do anything currently?

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,152 +1,236 @@
...
-    }
-    handleTouchStart (e) {
-        // Prevent this touch from becoming a click if menu is closed
-        if (!this.state.isOpen) {
-            e.preventDefault();
-            this.handleToggleOpenState();
+        }), []
+    );
+
+    const handleItemMouseEnter = useCallback(index => () => {
```
KManolov3, line 141: Can you leave a comment why this is needed?

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,152 +1,236 @@
...
-                className={classNames(styles.menuContainer, className, {
-                    [styles.expanded]: this.state.isOpen,
-                    [styles.forceHidden]: this.state.forceHide
-                })}
-                ref={this.setContainerRef}
-                onMouseEnter={this.handleToggleOpenState}
-                onMouseLeave={this.handleClosePopover}
+    }, [focusItem]);
+
+    const handleCoreButtonMouseEnter = useCallback(() => {
```
KManolov3: Would this be able to reuse some of the logic in `handleItemMouseEnter`?

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,152 +1,236 @@
...
+}) => {
+    const [forceHide, setForceHide] = useState(false);
+    const [isExpanded, setIsExpanded] = useState(false);
+    const ignoreNextFocusRef = useRef(false);
+
+    const closeTimeoutRef = useRef(null);
+    const mainTooltipId = useRef(`tooltip-${Math.random()}`).current;
+    const containerRef = useRef(null);
+    const buttonRef = useRef(null);
+    const itemRefs = useRef([]);
```
KManolov3, line 26: Nitpick: a comment why this is needed might also be useful

KManolov3 (approved):
Left a few minor comments, but otherwise the behavior looks very smooth to me now! 🙌  
There's a lot of logic "under the hood" which may end up being finicky, but hopefully this doesn't end up getting changed very often.

packages/scratch-gui/src/components/action-menu/action-menu.jsx
kbangelov: Hmm, it seems to be another issue that appears on firefox but not chrome. Upon reconsideration, I am not sure if **Escape** behavior is something we want at all here. I mean there are other intuitive ways to focus away as a keyboard or mouse user, such as **tab** or clicking outside of the menu. What do you think?

There is the "we want to be able to focus the core option" argument but the bottom one already duplicates that behavior on every use of action menu.

A compromise would be "focus on the main button on escape but keep the menu open, since we can still enter it with arrows". Collapsing seems a little unnatural to me now.

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -225,6 +224,7 @@ const ActionMenu = ({
                                 onClickItem(e);
                                 // timeout in case it loads something that loses the tooltip message
                                 // the exact delay number may need to adjust based on the time it takes to load
+                                lastFocusedItem.current = e.currentTarget;
```
KManolov3: Why is this between the comment and the setTimeout?

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -112,26 +125,30 @@ const ActionMenu = ({
...
             closeTimeoutRef.current = null;
+            lastFocusedItem.current = null;
         }, CLOSE_DELAY);
     }, []);
 
     useEffect(() => {
-        if (!isExpanded) {
+        if (isExpanded) {
+            refocusLastFocusedItem();
+            ReactTooltip.hide();
```
KManolov3: We are doing `hide` in both cases of the if

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -222,12 +196,10 @@ const ActionMenu = ({
                             
                             const handleClick = useCallback(e => {
```
KManolov3: Having a useCallback inside of a map is an antipattern - https://react.dev/reference/rules/rules-of-hooks#only-call-hooks-at-the-top-level. We should be able to extract this on top level.

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,152 +1,212 @@
...
+        }
+    }, []);
+   
+    useEffect(() => {
+        if (isExpanded) {
+            setTimeout(() => {
+                const focusedElement = document.activeElement;
+                focusedElement.blur();
+                focusItem(focusedElement);
+            }, CLOSE_DELAY * 1.5);
```
adzhindzhi: nitpick: We could probably extract this into a function, since we use the same logic twice?

adzhindzhi (approved):
Looks good to me! 🙌

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,152 +1,212 @@
...
+            item.focus();
+        }
+    }, []);
+
+    const refocusActiveElement = useCallback(() => {
+        setTimeout(() => {
+            const focusedElement = document.activeElement;
+            focusedElement.blur();
+            focusItem(focusedElement);
+        }, CLOSE_DELAY * 1.5);
```
KManolov3: 1.5 is a magic constant. Why 1.5 exactly - can we add a comment?

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,152 +1,212 @@
...
+        if (item) {
+            item.focus();
+        }
+    }, []);
+
+    const refocusActiveElement = useCallback(() => {
+        setTimeout(() => {
+            const focusedElement = document.activeElement;
+            focusedElement.blur();
+            focusItem(focusedElement);
```
KManolov3, line 38: `focusItem` should be a dependency (it doesn't have any dependencies itself currently, but that may change in the future)

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -1,152 +1,212 @@
...
+        }, CLOSE_DELAY * 1.5);
+    }, []);
+   
+    useEffect(() => {
```
KManolov3, line 45: Can we add comments in what usecases the different useEffects will be relevant

KManolov3 (approved):

packages/scratch-gui/src/components/action-menu/action-menu.jsx
```diff
@@ -36,9 +36,12 @@ const ActionMenu = ({
             const focusedElement = document.activeElement;
             focusedElement.blur();
             focusItem(focusedElement);
-        }, CLOSE_DELAY * 1.5);
-    }, []);
+        }, 500);
```
KManolov3, line 39: 500 is also a magic constant 😅 I think it was a bit better when it was connected with the CLOSE_DELAY. But we can update this later.

PR #434, opened 2026-02-03
Add `scratch-storage` to the monorepo

cwillisf:
### Proposed Changes

- Rebuild `build-monorepo.sh`, which rebuilt the whole monorepo fresh, as `add-to-monorepo.sh`, which adds a repository to the existing monorepo.
- Use `add-to-monorepo.sh` to add `scratch-storage`, then apply manual fixups.

### Reason for Changes

The `scratch-storage` package now uses `task-herder`, which means `scratch-editor` depends on `scratch-storage` _and_ `scratch-storage` depends on `scratch-editor`. I probably should have done this before f09c10292f43876b55b063b7f196d28afd034249, but we're here now :sweat_smile:

### Test Coverage

Existing tests in `scratch-storage` and in packages that depend on `scratch-storage` still pass.


cwillisf:
I'll rebase this on `develop` right before merging. Until then, Renovate will keep causing merge conflicts.

packages/scratch-storage/package.json
```diff
@@ -0,0 +1,59 @@
...
+  "homepage": "https://github.com/scratchfoundation/scratch-storage#readme",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/scratchfoundation/scratch-editor.git"
+  },
+  "license": "AGPL-3.0-only",
+  "main": "./dist/node/scratch-storage.js",
+  "browser": "./dist/web/scratch-storage.js",
+  "types": "./dist/types/index.d.ts",
+  "scripts": {
```
adzhindzhi, line 14: I'm not sure if that applies to `scratch-storage`, but other packages in the monorepo also include a `prepublishOnly` script that warns about publishing outside of CI.

packages/scratch-storage/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "test:unit": "jest \"test[\\\\/]unit\"",
+    "watch": "webpack --watch"
+  },
+  "tap": {
+    "check-coverage": false
+  },
+  "dependencies": {
+    "@babel/runtime": "^7.21.0",
+    "@scratch/task-herder": "13.7.4-svg",
+    "arraybuffer-loader": "^1.0.3",
```
adzhindzhi, line 33: There are a few packages where the version in `scratch-storage` differs from the versions used across the monorepo (e.g., `arraybuffer-loader`, `base64-js`). I wonder if Renovate will pick up on that and update them once this is merged, or if we'll have to handle it manually - assuming we want to handle it at all?

adzhindzhi (approved):
Looks good!!! 🙌

packages/scratch-storage/package.json
cwillisf: Yeah, Renovate should take care of this

packages/scratch-storage/package.json
cwillisf: Good catch! I've added that to `add-repo.sh` and then I ran it again.

PR #435, opened 2026-02-04
[UEPR-500] Moving costumes using the arrow keys changes tabs instead

kbangelov:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-500

### Proposed Changes

Artificially add an event listener for any clicks on the page so that clicking on canvas will remove the focus for tab (which previously it did not).

### Reason for Changes

Part of an accessibility initiative for Scratch.

### DIFF
--- a/packages/scratch-gui/src/components/gui/gui.jsx
+++ b/packages/scratch-gui/src/components/gui/gui.jsx
@@ -208,20 +208,6 @@ const GUIComponent = props => {
     }
 
     useEffect(() => {
-        const handleDocumentClick = () => {
-            // If any element is focused, blur it.
-            // Usually handled automatically but canvas clicks don't seem to blur previous focus.
-            if (document.activeElement !== document.body) {
-                document.activeElement.blur();
-            }
-        };
-        document.addEventListener('mousedown', handleDocumentClick);
-        return () => {
-            document.removeEventListener('mousedown', handleDocumentClick);
-        };
-    }, []);
-
-    useEffect(() => {
         if (props.platform) {
             // TODO: This uses the imported `setPlatform` directly,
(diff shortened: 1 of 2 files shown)


packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -207,6 +207,20 @@ const GUIComponent = props => {
         return <Box {...componentProps}>{children}</Box>;
     }
 
+    useEffect(() => {
+        const handleDocumentClick = () => {
+            // If any element is focused, blur it.
+            // Usually handled automatically but canvas clicks don't seem to blur previous focus.
+            if (document.activeElement) {
```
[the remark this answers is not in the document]
adzhindzhi: I don't think that's a valid comment, we're blurring the previously focused element, which will get blurred on click anyway (except when the `canvas` is clicked).

packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -207,6 +207,20 @@ const GUIComponent = props => {
         return <Box {...componentProps}>{children}</Box>;
     }
 
+    useEffect(() => {
+        const handleDocumentClick = () => {
```
adzhindzhi: Can we move this logic to the `CostumeTab`, since it only affects the Paint Editor? That way, we reduce the risk of it impacting other parts of the application in unexpected ways. We may want to move it back to the root level at some point, but this seems safer for now.

kbangelov: Yeah, fair.

adzhindzhi (approved):
Looks good to me!

PR #438, opened 2026-02-09
[UEPR-445] Topbar unit and integration tests

kbangelov:
### Resolves

Part of https://scratchfoundation.atlassian.net/browse/UEPR-445

### Proposed Changes

Unit and integration tests to ensure safety.

### Reason for Changes

### Test coverage

13 Unit tests for the hook and 17 Integration tests for the Settings, File and Edit menus


packages/scratch-gui/test/unit/hooks/use-menu-navigation.test.jsx
```diff
@@ -0,0 +1,338 @@
...
+    let item2;
+
+    beforeEach(() => {
+        jest.useFakeTimers();
+
+        store = configureStore({
+            reducer: {
+                locales: (state = {isRtl: false}) => state
+            }
+        });
```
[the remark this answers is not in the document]
kbangelov: It's a a deprecated method and this new one is the official recommendation. Yes, it's inconsistent and it seems to work the other way too, but I don't think it's a big deal. Perhaps I could change the others too?

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,293 @@
...
+    - Might need to change in case menus/submenus are moved around/reordered
+    - Added sleep time to ensure consistency between different keyboard events
+*/
+describe('Menu bar keyboard navigation', () => {
+    beforeAll(() => {
+        driver = getDriver();
+    });
+
+    afterAll(async () => {
+        await driver.executeScript('document.activeElement.blur()');
```
adzhindzhi: Why do we need to blur after all tests have completed?

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,293 @@
...
+        }
+    };
+
+    test('Tab focuses file menu on 3 clicks', async () => {
+        // Pressing tab 3 times should focus the file menu button
+        await clickKeys([Key.TAB, Key.TAB, Key.TAB]);
+
+        const activeElement = await driver.switchTo().activeElement();
+        expect(await activeElement.getAttribute('aria-label')).toBe('File menu');
+    });
```
adzhindzhi: Hm, this is a bit specific. I wonder if it would make more sense to check the "order of navigation", instead of the File Menu in particular. E.g. first we focus the logo, then Settings, then File menu, etc.? Or something like "Pressing Tab navigates through the menu items"?

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,293 @@
...
+        await driver.actions().sendKeys(key)
+            .perform();
+        await driver.sleep(SLEEP_TIME);
+    };
+
+    const clickKeys = async keys => {
+        for (const key of keys) {
+            await clickKey(key);
+        }
+    };
```
adzhindzhi: Maybe we could add these to the `SeleniumHelper`? If we want to write other tests for keyboard navigation, it'd be helpful to reuse them. It'd also be nice to add some error handling in case something goes wrong.

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,293 @@
...
+    });
+
+    beforeEach(async () => {
+        await loadUri(uri);
+    });
+
+    const clickKey = async key => {
+        await driver.actions().sendKeys(key)
+            .perform();
+        await driver.sleep(SLEEP_TIME);
```
adzhindzhi: Do we need to sleep after each key press? Or is there a particular issue you were avoiding that way? I wonder if we could avoid "sleeping" unless we definitely need to, to not increase the tests run time too much?

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,293 @@
...
+    });
+ 
+    test('Enter opens File menu', async () => {
+        const fileMenuButton = await findByXpath(FILE_MENU_XPATH);
+
+        expect(await fileMenuButton.getAttribute('aria-expanded')).toBe('false');
+
+        // Explicit keyboard focus
+        await driver.executeScript('arguments[0].focus()', fileMenuButton);
+        await clickKey(Key.ENTER);
```
adzhindzhi: I wonder if we could achieve the same by using `fileMenuButton.sendKeys(Key.ENTER)`, instead of manually focusing and then sending the key to the active element. Although for more complex test cases we'd still need to switch to the currently active element, and use `sendKeys` on it (e.g. for `ArrowUp`/`Down` navigation).

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,293 @@
...
+
+        const activeElement = await driver.switchTo().activeElement();
+        const text = await activeElement.getText();
+        expect(text).toBe('English');
+    });
+
+    test('Space opens the Language submenu', async () => {
+        const settingsMenuButton = await findByXpath(SETTINGS_MENU_XPATH);
+        await driver.executeScript('arguments[0].focus()', settingsMenuButton);
+        await clickKeys([Key.ENTER, Key.SPACE]);
```
adzhindzhi: Maybe we should test pressing `Space` twice? Because we've already tested the `Enter` interaction?

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,293 @@
...
+
+        await loadUri(uri);
+
+        const fileMenuButton2 = await findByXpath(FILE_MENU_XPATH);
+        await driver.executeScript('arguments[0].focus()', fileMenuButton2);
+        await clickKeys([Key.ENTER, Key.ARROW_DOWN, Key.TAB]);
+
+        const activeElement2 = await driver.switchTo().activeElement();
+        expect(await activeElement2.getAttribute('aria-label')).toBe('Edit menu');
+    });
```
adzhindzhi: The fact that we need to reload in the middle of the test makes me think that it'd make more sense to either divide these into two different test cases, e.g. `Tab closes File Menu` and `Tab closes Tab menu after navigation`, or test only the second one, since it also test whether "Tab closes the menu".

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,293 @@
...
+    test('Shift+Tab closes Edit menu and goes back to File menu', async () => {
+        const editMenuButton = await findByXpath(EDIT_MENU_XPATH);
+        await driver.executeScript('arguments[0].focus()', editMenuButton);
+        await clickKey(Key.ENTER);
+        await driver.actions()
+            .keyDown(Key.SHIFT)
+            .sendKeys(Key.TAB)
+            .keyUp(Key.SHIFT)
+            .perform();
+        await driver.sleep(SLEEP_TIME);
```
adzhindzhi, line 233: Why do we need the `sleep` here?

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
kbangelov: In most cases we do want it, I think. Some tests occasionally fail when sleep is 0 because of not having enough time to react. I just tested with and without sleep (50ms vs 0) and in both cases the tests took just under 20 seconds to complete. So it doesn't seem to really prolong the test time.

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
kbangelov: Well I added it anywhere and here the click is a little bit more specific (because of shift) so I can't use clickKey(s)

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
kbangelov: I probably meant it for **afterEach** but yeah, since we reload anyways we don't need it.

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
kbangelov: Perhaps the second part isn't different enough for its own test so I'll just remove it.

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
kbangelov: Yeah that's another variant, I suppose. This current way of doing it seems more consistent from a readability standpoint though, at least in my view. Although there's not much separating either variant for me.

KManolov3:
General question - how much execution time do these tests add to the existing test suites?

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,280 @@
...
+
+        expect(await fileMenuButton.getAttribute('aria-expanded')).toBe('true');
+    });
+
+    test('ArrowUp moves focus to the last item in file menu', async () => {
+        const fileMenuButton = await findByXpath(FILE_MENU_XPATH);
+        await driver.executeScript('arguments[0].focus()', fileMenuButton);
+        await clickKey(Key.ENTER);
+
+        // ArrowUp should go to last item
```
KManolov3: Nitpick: this comment is redundant as the test name already implies this is the expected behavior

kbangelov:
> General question - how much execution time do these tests add to the existing test suites?

18s for integration and 1s for unit.

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,280 @@
...
+
+    test('ArrowUp moves focus to the last item in edit menu', async () => {
+        const editMenuButton = await findByXpath(EDIT_MENU_XPATH);
+        await driver.executeScript('arguments[0].focus()', editMenuButton);
+        await clickKey(Key.ENTER);
+        const activeElement = await driver.switchTo().activeElement();
+        const text = await activeElement.getText();
+        expect(text).toBe('Restore');
+
+        // ArrowUp should go to last item
```
KManolov3: Same as last comment

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,280 @@
...
+        await clickKey(Key.ARROW_UP);
+
+        const activeElement2 = await driver.switchTo().activeElement();
+        const text2 = await activeElement2.getText();
+
+        expect(text2).toBe('Turn on Turbo Mode');
```
KManolov3, line 129: A general note: it's not a great practice to depend on text for selecting elements - because if the text changes, that will cascade into test changes that may have nothing to do with the text itself. I know there are some other instances of similar logic in the editor's integration tests (e.g. we have `findByText`). I also think that it'll be out of the scope of this task to implement a more resillient mechanism for selecting items (e.g. aria attributes) - just outlying some thoughts.

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,280 @@
...
+
+    test('ArrowDown twice moves focus back to the first item in edit menu', async () => {
+        const editMenuButton = await findByXpath(EDIT_MENU_XPATH);
+        await driver.executeScript('arguments[0].focus()', editMenuButton);
+        await clickKeys([Key.ENTER, Key.ARROW_DOWN, Key.ARROW_DOWN]);
+
+        const activeElement = await driver.switchTo().activeElement();
+        const text = await activeElement.getText();
+
+        expect(text).toBe('Restore');
```
KManolov3, line 140: Would it make sense for this to also check the length of the edit menu subitems, and make the checks based on that length (e.g. we expect two items, we are on the first item after pressing enter, and are on the first item again after pressing arrow down twice)

KManolov3: Similar logic may be able to be used for some of the other tests' comparisons - for example 'ArrowUp moves focus to the last item in edit menu'

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,280 @@
...
+    });
+
+    test('Enter opens the Language submenu', async () => {
```
KManolov3: `Enter opens the Language submenu` should probably omit the "Language" from its naming (and possibly its logic as well, if we can check whether the first item is a submenu). If it's quite a bit easier to check directly for the "Language" submenu, I am okay with leaving this check, but generally let's try to abstract ourselves from the concrete layout where possible  - the goal of these test cases is to check the menu tab navigation logic.

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,280 @@
...
+        const settingsMenuButton = await findByXpath(SETTINGS_MENU_XPATH);
+        await driver.executeScript('arguments[0].focus()', settingsMenuButton);
+        await clickKeys([Key.ENTER, Key.ARROW_RIGHT]);
+
+        const activeElement = await driver.switchTo().activeElement();
+        const text = await activeElement.getText();
+        expect(text).toBe('English');
+    });
+
+    test('ArrowLeft and Escape close the Language submenu', async () => {
```
KManolov3: Nitpick: given that the previous tests were very granular, it seems strange to combine two test cases in this one

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
KManolov3: Same comment about naming for other testcases referencing a concrete menu.

When you execute the tests, the name is formed from the describe clause + the test name. So this would be:
`Menu bar keyboard navigation Enter opens the Language submenu` which doesn't make sense as a normal sentence.

It might be better to have it as "can open submenu with Enter", so that the whole test becomes `Menu bar keyboard navigation can open submenu with Enter`

What do you think?

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,280 @@
...
+        const activeElement = await driver.switchTo().activeElement();
+        expect(await activeElement.getAttribute('aria-label')).toBe('Settings menu');
+        
+        await clickKey(Key.ENTER);
+        const activeElement2 = await driver.switchTo().activeElement();
+        expect(await activeElement2.getText()).toBe('Language');
+
+        await clickKey(Key.ARROW_DOWN);
+        const activeElement3 = await driver.switchTo().activeElement();
+        expect(await activeElement3.getText()).toBe('Color Mode');
```
KManolov3: I'd remove most of these intermediate checks, as they don't add much to the testcase, but make it much more prone to breaking if anything in the layout changes.

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,280 @@
...
+            .sendKeys(Key.TAB)
+            .keyUp(Key.SHIFT)
+            .perform();
+        await driver.sleep(SLEEP_TIME);
+
+        const activeElement = await driver.switchTo().activeElement();
+        expect(await activeElement.getAttribute('aria-label')).toBe('File menu');
+    });
+
+    test('Changing to RTL language should change ARROW_LEFT and ARROW_RIGHT behavior', async () => {
```
KManolov3: That's a useful test to have, thanks!

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,280 @@
...
+        const settingsMenuButton = await findByXpath(SETTINGS_MENU_XPATH);
+        await driver.executeScript('arguments[0].focus()', settingsMenuButton);
+        await clickKeys([Key.ENTER, Key.ARROW_RIGHT]);
+        const persianMenuItem = await findByXpath(
```
KManolov3, line 243: Any chance we can depend on aria labels here, as we do in the tests above?

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
```diff
@@ -0,0 +1,280 @@
...
+        await clickKey(Key.ENTER);
+
+        await clickKeys([Key.TAB, Key.TAB, Key.ENTER]);
+
+        await clickKey(Key.ARROW_LEFT);
+        const activeElement = await driver.switchTo().activeElement();
+        const text = await activeElement.getText();
+        expect(text).toBe('فارسی');
+        // Suspiciously fragile logic, might need to
+        // be altered in case of language option changes
```
KManolov3: I generally agree, though I think this might be one of the more resilient selectors 😄

packages/scratch-gui/test/unit/hooks/use-menu-navigation.test.jsx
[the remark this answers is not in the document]
KManolov3: I am okay with adding @reduxjs/toolkit since it's recommended to move onto it. Let's use it only here for now - I prefer us to migrate the rest of the codebase separately, as these PRs already carry a lot of changes with them

packages/scratch-gui/test/unit/hooks/use-menu-navigation.test.jsx
```diff
@@ -0,0 +1,338 @@
...
+
+    // Testing the following structure of the menu, where the button should not be misread as an item:
+    // <div data-menu-item-wrapper>
+    //  <button data-menu-item>Menu Button</button>
+    //  <div> - submenu
+    //   <li data-menu-item>Item 1</li>
+    //   <li data-menu-item>Item 2</li>
+    //  </div>
+    // <div>
+    test('handleOnOpen focuses first item inside submenu wrapper, not the button and then refocuses on close', () => {
```
KManolov3: -> `focuses first item inside submenu wrapper with handleOnOpen, skipping parent menu button` ?

packages/scratch-gui/test/unit/hooks/use-menu-navigation.test.jsx
```diff
@@ -0,0 +1,338 @@
...
+            menuContextMock.openInnerMenu.mockClear();
+            result.current.handleKeyDown({
+                key,
+                preventDefault: jest.fn(),
+                stopPropagation: jest.fn()
+            });
+            expect(buttonFocusSpy).not.toHaveBeenCalled();
+        });
+    });
+});
```
KManolov3, line 367: Let's also add a test checking that items without `data-menu-item` / with `data-menu-item` false are skipped over

KManolov3 (commented):
Great job! I really like how the tests have turned out, especially the unit ones. I think they will be very usefuil both for documentation and against regressions.

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
kbangelov: I feel like that would still be partly menu-specific because of how many times we press the arrow to loop. Do you think the current way is too high-test-maintenance?

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
kbangelov: I agree about the naming and remaining flexible with the title but isn't part of the integration tests's task to also test the concrete menus that we have so far (at least a few ones that are easiest to test). To me it seems normal that a test might need to change if "Language Menu" becomes not the first but the second submenu in settings or gets moved elsewhere. That shouldn't happen too often.

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
kbangelov: The menu items in language menu don't have **aria-label** (They could but they don't). So this is probably the way to find that language.

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
KManolov3: Yes, my concerns are more towards us having to make adjustments to the test every time we change something related to the menus - e.g. change the wording of a menu item. I agree that we'll always be dependent on the structure of the UI, since this is an integration test - but it's better to depend just on the structure, than on the text as well.

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
KManolov3: Yes, you're correct on both points - that integration tests should test that the concrete components visualise, interact in certain ways etc.., as well as that it shouldn't change too often.
But the goals of these tests in particular should be to test the menus' keyboard navigation behavior, and as such it's okay to abstract ourselves from the concrete menu implementation details

packages/scratch-gui/test/unit/hooks/use-menu-navigation.test.jsx
```diff
@@ -63,6 +63,35 @@ describe('useMenuNavigation', () => {
         expect(result.current.isExpanded()).toBe(false);
     });
 
+    test('skips over elements without data-menu-item or with data-menu-item="false"', () => {
+        const wrapperDiv = document.createElement('div');
+
+        const itemWithFalse = document.createElement('button');
```
KManolov3: nitpick: itemWithFalse isn't very descriptive as to what might be false

packages/scratch-gui/test/integration/menu-bar-keyboard-nav.test.js
kbangelov: Agreed. But is it not better to expect text to be "Restore", rather than copy the same BFS looking for subitems logic from the hook itself, as the first item being "Restore" or whatever will likely not change too often?

KManolov3: I am okay with leaving the current structure for this PR - realistically, we shouldn't be making changes to either the menu structure or labels very often.

KManolov3 (approved):

adzhindzhi (approved):

PR #442, opened 2026-02-12
[UEPR-488] Make Sprite selector items fully navigable via keyboard

kbangelov:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-488

### Proposed Changes

- Makes sprite item and delete button for it respond to "Enter" like they do on click.
- Improved semantics.
- Added default focus on delete button prompt open.
- Added better **aria-label** for sprite item that contains name and details.

### Reason for Changes

### DIFF
--- a/packages/scratch-gui/src/components/delete-button/delete-button.jsx
+++ b/packages/scratch-gui/src/components/delete-button/delete-button.jsx
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React, {useCallback} from 'react';
+import React from 'react';
 import classNames from 'classnames';
 
 import styles from './delete-button.css';
--- a/packages/scratch-gui/src/components/delete-button/delete-button.jsx
+++ b/packages/scratch-gui/src/components/delete-button/delete-button.jsx
@@ -5,39 +5,28 @@ import classNames from 'classnames';
 import styles from './delete-button.css';
 import deleteIcon from './icon--delete.svg';
 
-const DeleteButton = props => {
-
-    const handleKeyDown = useCallback(event => {
-        if (event.key === 'Enter' || event.key === ' ') {
-            props.onClick(event);
-        }
-    }, [props.onClick]);
-
-    return (
-        <button
-            aria-label="Delete"
-            className={classNames(
-                styles.deleteButton,
-                props.className
-            )}
-            tabIndex={props.tabIndex}
-            onClick={props.onClick}
-            onKeyDown={handleKeyDown}
+const DeleteButton = props => (
+    <button
+        aria-label="Delete"
+        className={classNames(
+            styles.deleteButton,
+            props.className
+        )}
+        tabIndex={props.tabIndex}
+        onClick={props.onClick}
+    >
+        <div
+            className={classNames(styles.deleteButtonVisible, {
+                [styles.deleteButtonClicked]: props.isConfirmationModalOpened
+            })}
         >
-            <div
-                className={classNames(styles.deleteButtonVisible, {
-                    [styles.deleteButtonClicked]: props.isConfirmationModalOpened
-                })}
-            >
-                <img
-                    className={styles.deleteIcon}
-                    src={deleteIcon}
-                />
-            </div>
-        </button>
-    );
-
-};
+            <img
+                className={styles.deleteIcon}
+                src={deleteIcon}
+            />
+        </div>
+    </button>
+);
 
 DeleteButton.propTypes = {
     className: PropTypes.string,
--- a/packages/scratch-gui/src/components/delete-button/delete-button.jsx
+++ b/packages/scratch-gui/src/components/delete-button/delete-button.jsx
@@ -5,30 +5,28 @@ import classNames from 'classnames';
 import styles from './delete-button.css';
 import deleteIcon from './icon--delete.svg';
 
-const DeleteButton = props => {
-    return (
-        <button
-            aria-label="Delete"
-            className={classNames(
-                styles.deleteButton,
-                props.className
-            )}
-            tabIndex={props.tabIndex}
-            onClick={props.onClick}
+const DeleteButton = props => (
+    <button
+        aria-label="Delete"
+        className={classNames(
+            styles.deleteButton,
+            props.className
+        )}
+        tabIndex={props.tabIndex}
+        onClick={props.onClick}
+    >
+        <div
+            className={classNames(styles.deleteButtonVisible, {
+                [styles.deleteButtonClicked]: props.isConfirmationModalOpened
+            })}
         >
-            <div
-                className={classNames(styles.deleteButtonVisible, {
-                    [styles.deleteButtonClicked]: props.isConfirmationModalOpened
-                })}
-            >
-                <img
-                    className={styles.deleteIcon}
-                    src={deleteIcon}
-                />
-            </div>
-        </button>
-    );
-};
+            <img
+                className={styles.deleteIcon}
+                src={deleteIcon}
+            />
+        </div>
+    </button>
+);
 
 DeleteButton.propTypes = {
     className: PropTypes.string,
--- a/packages/scratch-gui/src/components/delete-button/delete-button.jsx
+++ b/packages/scratch-gui/src/components/delete-button/delete-button.jsx
@@ -1,33 +1,43 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, {useCallback} from 'react';
 import classNames from 'classnames';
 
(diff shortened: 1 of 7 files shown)


packages/scratch-gui/src/components/sprite-selector-item/sprite-selector-item.css
```diff
@@ -20,6 +20,13 @@
     cursor: pointer;
 
     user-select: none;
+
+    /* removing default button styling */
```
kbangelov: since turning a **div** into a **button** been done quite a few times now, surrounding semantics, perhaps we should consider a global css style that we can apply in such cases, since code seems to be essentially copy-pasted here.

packages/scratch-gui/src/components/sprite-selector-item/sprite-selector-item.jsx
```diff
@@ -20,24 +20,29 @@ const SpriteSelectorItem = props => {
...
                         [styles.isSelected]: props.selected
                     })}
                     onClick={props.onClick}
                     onMouseEnter={props.onMouseEnter}
                     onMouseLeave={props.onMouseLeave}
                     onMouseDown={props.onMouseDown}
+                    onKeyDown={handleKeyDown}
                     onTouchStart={props.onMouseDown}
                     ref={props.componentRef}
                 >
```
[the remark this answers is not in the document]
kbangelov: That is a good point, perhaps I should restructure in a way that keeps the design the same? Let's first discuss if we want the outer wrapper of the sprite item to actually be a button.

packages/scratch-gui/src/components/sprite-selector-item/sprite-selector-item.jsx
```diff
@@ -20,24 +20,29 @@ const SpriteSelectorItem = props => {
...
     return (
         <ContextMenu.Root modal={false}>
             <ContextMenu.Trigger
                 disabled={props.preventContextMenu}
                 asChild
             >
-                <div
-                    role="button"
-                    tabIndex={0}
+                <button
```
kbangelov: I'm also thinking I'm not sure if we want this to be a button. Visually it's not that clear if we want this. Any other ideas?

packages/scratch-gui/src/components/delete-button/delete-button.jsx
```diff
@@ -1,33 +1,43 @@
...
+        <button
+            aria-label="Delete"
+            className={classNames(
+                styles.deleteButton,
+                props.className
+            )}
+            tabIndex={props.tabIndex}
+            onClick={props.onClick}
+            onKeyDown={handleKeyDown}
         >
```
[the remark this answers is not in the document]
adzhindzhi: I agree with the bot that pressing Enter on a `<button>` already works out of the box. `Space` wouldn't work this case, but I think we should stick to the native keyboard behavior of the `button` rather than introducing custom logic. I'm also not sure whether we actually want the `button` to be triggered via `Space`, or how common that interaction would be in this context, so it seems fine to me to stick to the default behavior.

packages/scratch-gui/src/components/sprite-selector-item/sprite-selector-item.jsx
adzhindzhi: We already had that discussion for the [semantic HTML PR](https://github.com/scratchfoundation/scratch-editor/pull/408/changes/BASE..04253d923eea2b1f71cfad128de95ad8fa9db2bb#r2664551395). We intentionally left this as a `div role="button"` with the note that we'd have to handle `onKeyDown` manually to make it fully interactive. I think we should stick to that approach so that we don't render `<button>` inside a `<button>`.

packages/scratch-gui/src/components/delete-confirmation-prompt/delete-confirmation-prompt.jsx
adzhindzhi: Hm, I'm not entirely convinced that we need this. Since we're using a modal, focus is already trapped inside it. The first focusable element would receive focus when the user presses `Tab` anyway. Automatically focusing the button on mount seems fine, but it does differ from the default modal behavior, and I'm not sure we gain a clear accessibility benefit from doing so. Do we actually improve accessibility by forcing focus to the primary button immediately? Maybe we should run this by Tereza to see if this makes sense from a UX perspective.

If we do decide this behavior is required, we could probably simplify the implementation by using the modal's `onAfterOpen` prop instead:
```
onAfterOpen={() => {
  if (yesButtonRef.current) {
    yesButtonRef.current.focus();
  }
}}
```

That would avoid the `useEffect` + `requestAnimationFrame` and keep the focus logic closer to the modal lifecycle.

kbangelov: I agree about the built-in prop.

And yes, it does change our behaviour by highlighting "Yes" on the prompt but that's what the user presumably wants to do, I think saving one click if navigating with keyboard seems logical.

adzhindzhi: Both behaviors make sense. Usually, I'd advocate for keeping the default behavior unless it really makes sense not to, and using as little custom logic as possible to avoid introducing unwanted changes. This seems like a low-risk change, but unless it really makes sense from a UX perspective, I'd say let's keep the default behavior :) I, personally, as a non-keyboard user would find it weird that the `Yes` button is automatically focused when I open the modal.

packages/scratch-gui/src/components/sprite-selector-item/sprite-selector-item.jsx
```diff
@@ -20,7 +21,13 @@ const SpriteSelectorItem = props => {
         window.addEventListener('resize', handleResize);
         return () => window.removeEventListener('resize', handleResize);
     }, []);
-    
+
+    const handleKeyDown = useCallback(event => {
+        if (event.key === KEY.ENTER && event.currentTarget === event.target) {
```
adzhindzhi, line 26: Out of curiosity, why do we need to check `event.target` and `event.currentTarget`?

packages/scratch-gui/src/components/sprite-selector-item/sprite-selector-item.jsx
```diff
@@ -31,10 +38,12 @@ const SpriteSelectorItem = props => {
                 <div
                     role="button"
                     tabIndex={0}
+                    aria-label={`${props.name}${props.details ? ` (${props.details})` : ''}`}
```
adzhindzhi: nitpick:
```
aria-label={
  props.details ? `${props.name} (${props.details})` : props.name
}
```

Also, wouldn't screen readers read the visible text anyway? Why do we need to add the `aria-label` manually?

packages/scratch-gui/src/components/sprite-selector-item/sprite-selector-item.jsx
kbangelov: Because of a button inside button bug, where enter doesn't otherwise apply for the delete icon but for the sprite item instead.

packages/scratch-gui/src/components/sprite-selector-item/sprite-selector-item.jsx
kbangelov: It might make little difference for a good screen reader, still I think it's worth it to have a little more control over what gets read in such a situation, adding some punctuation as well.

packages/scratch-gui/src/components/delete-confirmation-prompt/delete-confirmation-prompt.jsx
```diff
@@ -1,5 +1,5 @@
 import {defineMessages, FormattedMessage, useIntl} from 'react-intl';
-import React from 'react';
+import React, {useCallback} from 'react';
```
adzhindzhi: Can you remove the unused import, please 🙏

packages/scratch-gui/src/components/delete-confirmation-prompt/delete-confirmation-prompt.jsx
```diff
@@ -91,8 +91,6 @@ const DeleteConfirmationPrompt = ({
 
     return (<ReactModal
         isOpen
-        // We have to inline the styles, since a part
-        // of them are dynamically generated
```
adzhindzhi, line 95: Let's bring this comment back, it seems like it still makes sense?

adzhindzhi (approved):
Left two minor comments, but otherwise looks good to me! 🙌

adzhindzhi (approved):
Looks good! 🙌

PR #443, opened 2026-02-12
[UEPR-494] Text field not automatically focused

adzhindzhi:
### Resolves

[UEPR-494](https://scratchfoundation.atlassian.net/browse/UEPR-494)

### Proposed Changes

- Only attempt to blur if the currently focused element is actually the tab, rather than always blurring

### Reason for Changes

- The text field gains focus before our attempt to blur focus from the tabs, which results in users not being able to type right away

[UEPR-494]: https://scratchfoundation.atlassian.net/browse/UEPR-494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/scratch-gui/src/containers/costume-tab.jsx
+++ b/packages/scratch-gui/src/containers/costume-tab.jsx
@@ -113,7 +113,9 @@ class CostumeTab extends React.Component {
     }
     componentDidMount () {
         this.handleDocumentClick = () => {
-            if (document.activeElement !== document.body) {
+            // If the costume tab is focused and the user clicks outside of it, unfocus the costume tab.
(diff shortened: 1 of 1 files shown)


KManolov3 (approved):
LGTM, it might be good to have this on the other tabs as well, but this should be the lowest risk change-set for a hotfix.

cwillisf (approved):
I agree with both parts of what @KManolov3 said :)
I'm tempted to suggest moving this to a base class for tabs, but maybe that's my C++ background talking...
This seems good as a specific bug fix, though. Thanks!

PR #444, opened 2026-02-12
power operator block `a^b`

### LINKED ISSUE #426: Add an exponent block please.
The new "online?" block added to Scratch proved that Scratch IS willing to add new blocks.

So can we _please_ have an exponent block?
There are countless reasons for why it should be added, which you can find if you read through [this topic](https://scratch.mit.edu/discuss/topic/2860/)

It's a block that has no particularly easy or simple workarounds, and it would be amazing if it could be added, or the suggestion at least considered instead of ignored.

AeroKoder:
See also https://github.com/scratchfoundation/scratch-blocks/pull/3462

### Resolves

#426 

### Proposed Changes

Add a new block to the Operator category to raise a given number to a power of another given number.

### Reason for Changes

This is a highly requested block by many Scratchers and Scratch Team members (I have chatted with some [[1](https://scratch.mit.edu/users/algorithmar/#comments-402360195),[2](https://scratch.mit.edu/users/glowinday/#comments-402404181)], and they agree that it should exist in Scratch). There also aren't any accurate workarounds, besides ones that are crazy inefficient, making this block even more important to have.

It has been discussed in multiple topics and issues, and they always end up with tons of upvotes and supports.

The one thing I would like to discuss is whether, zero to the power of zero, should return `1` or `NaN`. Currently, it returns `1`, as that is what the native JavaScript operator returns. This might confuse younger Scratchers, but then again, I don't know if `NaN` is such a good idea either, since it acts as zero in scratch, making some math equations with the power operator return an incorrect result.

### Test Coverage

Tested locally:
scratch-power-block-tested-locally

### DIFF
--- a/packages/scratch-vm/src/blocks/scratch3_operators.js
+++ b/packages/scratch-vm/src/blocks/scratch3_operators.js
@@ -55,7 +55,7 @@ class Scratch3OperatorsBlocks {
     }
 
     power (args) {
-        return Cast.toNumber(args.NUM1) ** Cast.toNumber(args.NUM2);
+        return Math.pow(Cast.toNumber(args.NUM1), Cast.toNumber(args.NUM2));
     }
 
     lt (args) {
--- a/packages/scratch-vm/src/blocks/scratch3_operators.js
+++ b/packages/scratch-vm/src/blocks/scratch3_operators.js
@@ -20,6 +20,7 @@ class Scratch3OperatorsBlocks {
             operator_subtract: this.subtract,
             operator_multiply: this.multiply,
             operator_divide: this.divide,
+            operator_power: this.power,
             operator_lt: this.lt,
             operator_equals: this.equals,
             operator_gt: this.gt,
@@ -51,6 +52,10 @@ class Scratch3OperatorsBlocks {
 
     divide (args) {
         return Cast.toNumber(args.NUM1) / Cast.toNumber(args.NUM2);
+    }
+
+    power (args) {
+        return Cast.toNumber(args.NUM1) ** Cast.toNumber(args.NUM2);
     }
 
     lt (args) {
--- a/packages/scratch-gui/src/lib/make-toolbox-xml.js
+++ b/packages/scratch-gui/src/lib/make-toolbox-xml.js
@@ -579,6 +579,18 @@ const operators = function (isInitialSetup, isStage, targetId, colors) {
                 </shadow>
             </value>
         </block>
+        <block type="operator_power">
+            <value name="NUM1">
+                <shadow type="math_number">
+                    <field name="NUM"/>
+                </shadow>
+            </value>
+            <value name="NUM2">
+                <shadow type="math_number">
+                    <field name="NUM"/>
+                </shadow>
+            </value>
+        </block>
         ${blockSeparator}
         <block type="operator_random">
             <value name="FROM">


AeroKoder:

redspacecat:
this would be great to add!
please merge this PR, scratch team.

Jab11n:
This has been highly requested on the Scratch Forums for years. It would be a major long-awaited addition to the Scratch editor should it be merged. In fact, [it was first suggested in 2013](https://scratch.mit.edu/discuss/topic/2860/). Please merge it!

Oren-Lindsey:
This is a great addition, I think a lot of people would appreciate it.

OceanIsEndless:
There are so many amazing projects that could be created with this, but are currently too tedious to do with workarounds (even the best ones have imprecision and are just too complicated). Scratch has its fair share of beginner and advanced features, and merging this costs nothing... please add this block!

nimeratus:
I think the block looks a bit strange in the topmost group - all other block groups have at most 4 blocks in them.

I've seen a suggestion that I think puts the block to a better position: https://scratch.mit.edu/discuss/post/8998382/

AeroKoder:
> I think the block looks a bit strange in the topmost group - all other block groups have at most 4 blocks in them.
> 
> I've seen a suggestion that I think puts the block to a better position: https://scratch.mit.edu/discuss/post/8998382/

I partially disagree with this, the fact that all of the other block categories don’t have over 4 blocks in their first group, makes me want to move the block down to right above the `mod` block. But since the power operator is more related to the multiply operator than to the mod operator, I think it should stay where I originally had it.

@cwillisf What’s your thoughts on this?

yuri-kiss:
@cwillisf can you please look at this ?, you guys recently added an "is online?" boolean so its not like far to the right of a suggestion, plus this one wont require any translations which is a plus :3

yuri-kiss:
@AeroKoder 

> The one thing I would like to discuss is whether, zero to the power of zero, should return 1 or NaN. Currently, it returns 1, as that is what the native JavaScript operator returns. This might confuse younger Scratchers, but then again, I don't know if NaN is such a good idea either, since it acts as zero in scratch, making some math equations with the power operator return an incorrect result.

you do realize you can already get NaN with the division operator right?

image

Besides that its on purpose, see https://en.wikipedia.org/wiki/Zero_to_the_power_of_zero

packages/scratch-vm/src/blocks/scratch3_operators.js
```diff
@@ -53,6 +54,10 @@ class Scratch3OperatorsBlocks {
         return Cast.toNumber(args.NUM1) / Cast.toNumber(args.NUM2);
     }
 
+    power (args) {
+        return Cast.toNumber(args.NUM1) ** Cast.toNumber(args.NUM2);
```
yuri-kiss: half joke, but imo you should consider using Math.pow instead, its slightly faster

yuri-kiss: dunno why this showed as a review, just wanted to make a comment on the line

AeroKoder: Interesting, I thought they would be the same. But I guess `Math.pow` is more efficient, do to the fact that `**` works with `BigInt`s, and `Math.pow` does not. Thanks for bringing that up, I'll switch it to `Math.pow` right now.

supervoidcoder:
> The one thing I would like to discuss is whether, zero to the power of zero, should return `1` or `NaN`. Currently, it returns `1`,

This is how it works in regular math anyways, it should be 1 to be consistent with mathematics. When kids get to Algebra 1 or 2 they will be taught that any number to the power of 0 is 1. And if this is taught earlier from the behavior of this block, then it will be learned even earlier, which, is, well, a good thing

supervoidcoder:
> > I think the block looks a bit strange in the topmost group - all other block groups have at most 4 blocks in them.
> > I've seen a suggestion that I think puts the block to a better position: https://scratch.mit.edu/discuss/post/8998382/
> 
> I partially disagree with this, the fact that all of the other block categories don’t have over 4 blocks in their first group, makes me want to move the block down to right above the `mod` block. But since the power operator is more related to the multiply operator than to the mod operator, I think it should stay where I originally had it.
> 
> @cwillisf What’s your thoughts on this?

I'm not @cwillisf but I can give my opinion. When new blocks are added to the pallette I would greatly like them to maintain muscle memory with the positions of where they are dragged from. By putting it at the end, you maintain muscle memory by putting it in an intuitive spot rather than putting it somewhere else which might not only make the user grab the wrong block, be confused, or be annoyed

rosics-code (approved):
LGTM!

Jab11n (approved):

PR #445, opened 2026-02-13
[UEPR-487] Make the Sound Editor fully navigable via keyboard

kbangelov:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-487

### Proposed Changes

Improved accessibility
- Disabled some buttons, preventing them from being focused when disabled.
- Improved outline of play button.
- Removed unnecessary **tabIndex=1** logic.
- Added aria-labels instead of **title** in buttons because it will be picked up by more screen-readers.
- Updated tests and snapshots to adjust to new attributes.

### Reason for Changes

### Test Coverage

Tests were already written for the Editor.

### DIFF
--- a/packages/scratch-gui/src/components/sound-editor/sound-editor.jsx
+++ b/packages/scratch-gui/src/components/sound-editor/sound-editor.jsx
@@ -157,6 +157,7 @@ const SoundEditor = props => {
                     <div className={styles.buttonGroup}>
                         <button
                             aria-label={intl.formatMessage(messages.undo)}
+                            title={intl.formatMessage(messages.undo)}
                             className={styles.button}
                             disabled={!props.canUndo}
                             onClick={props.onUndo}
@@ -171,6 +172,7 @@ const SoundEditor = props => {
                             className={styles.button}
                             disabled={!props.canRedo}
                             aria-label={intl.formatMessage(messages.redo)}
+                            title={intl.formatMessage(messages.redo)}
                             onClick={props.onRedo}
                         >
                             <img
@@ -233,6 +235,7 @@ const SoundEditor = props => {
                         <button
                             className={classNames(styles.roundButton, styles.stopButton)}
                             aria-label={intl.formatMessage(messages.stop)}
+                            title={intl.formatMessage(messages.stop)}
                             onClick={props.onStop}
                         >
                             <img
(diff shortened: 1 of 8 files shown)


packages/scratch-gui/src/components/sound-editor/sound-editor.jsx
```diff
@@ -149,18 +149,16 @@ const SoundEditor = props => {
...
                             value={props.name}
                             onSubmit={props.onChangeName}
                         />
                     </Label>
                     <div className={styles.buttonGroup}>
                         <button
+                            aria-label={intl.formatMessage(messages.undo)}
                             className={styles.button}
                             disabled={!props.canUndo}
-                            title={intl.formatMessage(messages.undo)}
```
adzhindzhi, line 163: Just curious, is there a reason why you decided to switch from `title` to `aria-label`? Is there a benefit in keeping both? I think there might be, since `title` also provides a tooltip for sighted users, and since these buttons are icon only, it might be nice to keep it.

adzhindzhi (commented):
I left one small comment, but otherwise it looks good!

packages/scratch-gui/src/components/sound-editor/sound-editor.jsx
kbangelov: As far as screen readers go, I think aria-label might be picked up a little more. Also could you clarify how the title attribute provides a tooltip? Not sure what you mean by that.

adzhindzhi: Yep, when you hover on an element with a `title` attribute for a few seconds, a tooltip gets displayed, e.g.:
image

kbangelov: Did not know that. Seems like a serious drawback, so I'll bring it back, leaving both to make ensure accessibility, since from what I read, some screen readers might not pick up **title** alone.

packages/scratch-gui/test/integration/sounds.test.js
```diff
@@ -49,7 +49,7 @@ describe('Working with sounds', () => {
         await clickText('Chomp'); // Should close the modal, then click the sounds in the selector
         await findByXpath("//input[@value='Chomp']"); // Should show editor for new sound
 
-        await clickXpath('//button[@title="Play"]');
+        await clickXpath('//button[@aria-label="Play"]');
```
adzhindzhi: Can we revert the changes to the tests to continue using `title`, rather than `aria-label`? And update the snapshots also? 🙏

adzhindzhi (approved):
Looks good to me! 🙌

PR #446, opened 2026-02-13
[UEPR-445] Ensure topbar is navigable via tab navigation

kbangelov:
### Aims to resolve UEPR-445

https://scratchfoundation.atlassian.net/browse/UEPR-445

### Proposed Changes

1. Made elements focusable
2. Made a context (**menu-ref-context**) that takes care of tracking which menus and submenus are currently open via their refs. All of the logic with the "**isOpen**" type of props has been replaced for these menus.
3. Made a **use-menu-navigation** hook to be used by the menus. Handles common key pressing, opening, closing logic and more. Uses customized BFS logic for HTML components

### To be done additionally concerning nav bar accessibility - in other PRs?
- Choosing a **language** option does not focus on **Settings** after. Also not for the other options either, but that's because not closing after selecting an option has been their behavior so far. Perhaps another issue could be made that clears whether they should or not.
- Login Menu doesn't navigate properly (close on tab for instance) because it's written in www
- **WARNING** Because of the improper structure of nesting menu inside button we have a button inside of button with **Settings Menu** and **Language Menu / Preference Menu**, causing a warning in console.
- Focus is not too pretty on some menu items.
- Write tests for navigation bar

### To be discussed additionally
- The behavior of using screen reader and clicking with cursor leads to improper reads, as opposed to just navigating with keyboard. But I don't know if that's a big issue.


adzhindzhi (approved):
This is a duplicate of [https://github.com/scratchfoundation/scratch-editor/pull/403](url) for merging into `release/UEPR-297-accessibility-improvements`. The original PR was already reviewed, approved and merged.

PR #450, opened 2026-02-19
[UEPR-516] Debug modal accessibility

kbangelov:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-516

### Proposed Changes

- Added keyboard navigation for the modal slides
- Made next slide and previous slide buttons focusable

### Reason for Changes

Part of Accessibility initiative for Scratch.

### DIFF
--- a/packages/scratch-gui/src/components/debug-modal/debug-modal.css
+++ b/packages/scratch-gui/src/components/debug-modal/debug-modal.css
@@ -160,3 +160,15 @@
         margin: 5px;
     }
 }
+
+/* TO DO: export this class for reuse */
+.button-style-remover {
+    background: none;
+    border: none;
+    padding: 0;
+    margin: 0;
+    font: inherit;
+    color: inherit;
+    cursor: pointer;
+    text-align: inherit;
+}
--- a/packages/scratch-gui/src/components/debug-modal/debug-modal.css
+++ b/packages/scratch-gui/src/components/debug-modal/debug-modal.css
@@ -99,6 +99,7 @@
 
     .topic-item {
         display: flex;
+        width: 100%;
         gap: 8px;
         align-items: center;
         padding: 8px;
@@ -156,7 +157,8 @@
         margin-top: 20px;
     }
 
-    button {
+    .modal-header button,
+    .navigation-buttons button {
         margin: 5px;
     }
 }
--- a/packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
+++ b/packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
@@ -31,7 +31,6 @@ const logTopicChange = topicIndex => {
 
 const DebugModal = ({isOpen, onClose = () => {}}) => {
     const [selectedTopicIndex, setSelectedTopicIndex] = useState(0);
-    const slideRefs = React.useMemo(() => sections.map(() => React.createRef()), []);
 
     // Preload images
     useEffect(() => {
@@ -70,27 +69,30 @@ const DebugModal = ({isOpen, onClose = () => {}}) => {
         if ([KEY.ARROW_LEFT, KEY.ARROW_UP, KEY.ARROW_RIGHT, KEY.ARROW_DOWN].includes(e.key)) {
             e.preventDefault();
 
-            let nextIndex;
-            if (e.key === KEY.ARROW_LEFT || e.key === KEY.ARROW_UP) {
-                nextIndex = selectedTopicIndex > 0 ? selectedTopicIndex - 1 : sections.length - 1;
-            } else {
-                nextIndex = selectedTopicIndex < sections.length - 1 ? selectedTopicIndex + 1 : 0;
-            }
+            setSelectedTopicIndex(prev => {
+                let nextIndex = prev;
 
-            if (slideRefs[nextIndex] && slideRefs[nextIndex].current) {
-                slideRefs[nextIndex].current.focus();
-            }
+                if ((e.key === KEY.ARROW_LEFT || e.key === KEY.ARROW_UP) && prev > 0) {
+                    nextIndex = prev - 1;
+                } else if ((e.key === KEY.ARROW_RIGHT || e.key === KEY.ARROW_DOWN) &&
+                    prev < sections.length - 1) {
+                    nextIndex = prev + 1;
+                }
 
-            setSelectedTopicIndex(nextIndex);
-            logTopicChange(nextIndex);
+                if (nextIndex !== prev) {
+                    logTopicChange(nextIndex);
+                }
+
+                return nextIndex;
+            });
         }
-    }, [selectedTopicIndex]);
+    }, []);
 
     useEffect(() => {
         if (!isOpen) return;
 
-        window.addEventListener('keydown', handleKeyDownSlides);
-        return () => window.removeEventListener('keydown', handleKeyDownSlides);
+        document.addEventListener('keydown', handleKeyDownSlides);
+        return () => document.removeEventListener('keydown', handleKeyDownSlides);
     }, [isOpen, handleKeyDownSlides]);
 
     const handleKeyDownPrevious = useCallback(e => {
@@ -153,9 +155,7 @@ const DebugModal = ({isOpen, onClose = () => {}}) => {
                             })}
                             // eslint-disable-next-line react/jsx-no-bind
                             onClick={() => handleTopicSelect(index)}
-                            tabIndex={-1}
                             role="button"
-                            ref={slideRefs[index]}
                         >
                             <div className={styles.debugIcon}>
                                 <img
--- a/packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
+++ b/packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
@@ -96,13 +96,13 @@ const DebugModal = ({isOpen, onClose = () => {}}) => {
     }, [isOpen, handleKeyDownSlides]);
 
     const handleKeyDownPrevious = useCallback(e => {
-        if (e.key === KEY.ENTER) {
+        if (e.key === KEY.SPACE) {
             handlePrevious();
         }
     }, [handlePrevious]);
 
     const handleKeyDownNext = useCallback(e => {
-        if (e.key === KEY.ENTER) {
+        if (e.key === KEY.SPACE) {
             handleNext();
         }
     }, [handleNext]);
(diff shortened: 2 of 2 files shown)


packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
```diff
@@ -64,6 +66,45 @@ const DebugModal = ({isOpen, onClose = () => {}}) => {
         onClose();
     }, [onClose]);
 
+    const handleKeyDownSlides = useCallback(e => {
+        if ([KEY.ARROW_LEFT, KEY.ARROW_UP, KEY.ARROW_RIGHT, KEY.ARROW_DOWN].includes(e.key)) {
+            e.preventDefault();
+
+            let nextIndex;
+            if (e.key === KEY.ARROW_LEFT || e.key === KEY.ARROW_UP) {
+                nextIndex = selectedTopicIndex > 0 ? selectedTopicIndex - 1 : sections.length - 1;
```
kbangelov: It is a bit 50/50 to me whether we should add the loop logic. On one hand, we could. On the other, the buttons so far don't suggest looping. But that's also a thing we can change if we wanted to.

kbangelov (commented):
Also, the close button's focus doesn't look the prettiest. Its behavior was actually not changed at all in this PR, but maybe it's still worth fixing in terms of styling?

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
adzhindzhi: Good question, I guess we can run it by Tereza from a UX perspective. I also wonder if `ARROW_LEFT`/`ARROW_RIGHT` should change the selection. Probably yes, it seems convenient?

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
```diff
@@ -64,6 +66,45 @@ const DebugModal = ({isOpen, onClose = () => {}}) => {
...
+            }
+
+            if (slideRefs[nextIndex] && slideRefs[nextIndex].current) {
+                slideRefs[nextIndex].current.focus();
+            }
+
+            setSelectedTopicIndex(nextIndex);
+            logTopicChange(nextIndex);
+        }
+    }, [selectedTopicIndex]);
```
adzhindzhi: Should `setSelectedTopicIndex` also be part of the deps array?

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
```diff
@@ -64,6 +66,45 @@ const DebugModal = ({isOpen, onClose = () => {}}) => {
...
+
+    useEffect(() => {
+        if (!isOpen) return;
+
+        window.addEventListener('keydown', handleKeyDownSlides);
+        return () => window.removeEventListener('keydown', handleKeyDownSlides);
+    }, [isOpen, handleKeyDownSlides]);
```
adzhindzhi, line 106: What I find slightly unfortunate about the current implementation is that the event listener will be re-attached on every topic change, rather than attaching it once when the dialog opens. And the reason that would happen is because in `handleKeyDownSlides` we rely on the `selectedTopicIndex`. Maybe we could do something like:
```
 setSelectedTopicIndex(prev => {
                let nextIndex;
                if (e.key === KEY.ARROW_LEFT || e.key === KEY.ARROW_UP) {
                    // logic here based on prev
                    logTopicChange(nextIndex);
            });
```

for `handleKeyDownSlides`, to avoid the `selectedTopicIndex` dependency and avoid re-attaching the event listener?

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
```diff
@@ -112,6 +153,9 @@ const DebugModal = ({isOpen, onClose = () => {}}) => {
                             })}
                             // eslint-disable-next-line react/jsx-no-bind
                             onClick={() => handleTopicSelect(index)}
+                            tabIndex={-1}
```
adzhindzhi, line 159: Do you need this? This is un-focusable through `Tab` navigation by default, isn't it?

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
```diff
@@ -64,6 +66,45 @@ const DebugModal = ({isOpen, onClose = () => {}}) => {
...
+
+            setSelectedTopicIndex(nextIndex);
+            logTopicChange(nextIndex);
+        }
+    }, [selectedTopicIndex]);
+
+    useEffect(() => {
+        if (!isOpen) return;
+
+        window.addEventListener('keydown', handleKeyDownSlides);
```
adzhindzhi: nitpick: I wonder if we should attach the event listener to the modal element (or at least `document`) to make it more DOM-scoped.

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
kbangelov: Tereza said not to loop it with keys so I am changing that.

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
kbangelov: No, it was part of logic for manually focusing but we don't need to do that at all, so I removed it.

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
kbangelov: **<ReactModal>** doesn't seem to pick up **onKeyDown** so I'll just change to document.

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
```diff
@@ -150,6 +194,9 @@ const DebugModal = ({isOpen, onClose = () => {}}) => {
                             className={classNames(styles.previousIcon, {
                                 [styles.hidden]: selectedTopicIndex === 0
                             })}
+                            tabIndex={0}
+                            role="button"
```
KManolov3: Can we rather have the `img` tags be wrapped in buttons, instead of trying to simulate button behavior?

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
```diff
@@ -64,6 +65,48 @@ const DebugModal = ({isOpen, onClose = () => {}}) => {
...
+
+    useEffect(() => {
+        if (!isOpen) return;
+
+        document.addEventListener('keydown', handleKeyDownSlides);
+        return () => document.removeEventListener('keydown', handleKeyDownSlides);
+    }, [isOpen, handleKeyDownSlides]);
+
+    const handleKeyDownPrevious = useCallback(e => {
+        if (e.key === KEY.ENTER) {
```
KManolov3: Maybe it makes sense to allow interaction with `Space` as well.

KManolov3: Do we really need to specify this explicitly? If we used a `<button` wouldn't ENTER and SPACE work out of the box?

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
```diff
@@ -64,6 +65,48 @@ const DebugModal = ({isOpen, onClose = () => {}}) => {
...
+
+                if ((e.key === KEY.ARROW_LEFT || e.key === KEY.ARROW_UP) && prev > 0) {
+                    nextIndex = prev - 1;
+                } else if ((e.key === KEY.ARROW_RIGHT || e.key === KEY.ARROW_DOWN) &&
+                    prev < sections.length - 1) {
+                    nextIndex = prev + 1;
+                }
+
+                if (nextIndex !== prev) {
+                    logTopicChange(nextIndex);
```
KManolov3, line 93: We should've removed these logs a while ago... as they are currently, they just add bloat to Google Analytics. It's fine to leave it as-is for now.

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
```diff
@@ -64,6 +65,48 @@ const DebugModal = ({isOpen, onClose = () => {}}) => {
...
+                return nextIndex;
+            });
+        }
+    }, []);
+
+    useEffect(() => {
+        if (!isOpen) return;
+
+        document.addEventListener('keydown', handleKeyDownSlides);
+        return () => document.removeEventListener('keydown', handleKeyDownSlides);
```
KManolov3, line 105: Shouldn't this eventListener also be removed when `isOpen` becomes false?

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
kbangelov: It does seem like that kind of button.

kbangelov: Enter, yes. Space, no. So i'll just keep Space for now.

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
kbangelov: seems identical with the condition of unmounting to me?

packages/scratch-gui/src/components/debug-modal/debug-modal.css
```diff
@@ -160,3 +160,15 @@
         margin: 5px;
     }
 }
+
+/* TO DO: export this class for reuse */
+.button-style-remover {
```
KManolov3, line 167: Why is this needed?

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
KManolov3: In `gui.jsx` `DebugModal` is always being rendered - whether it's content is shown however is being controlled from outside by the `isOpen` flag. So we won't go into the unmounting logic which removes the event listener. 
Feel free to test it locally, and see whether we end up with multiple registered event listeners.

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
```diff
@@ -112,6 +155,7 @@ const DebugModal = ({isOpen, onClose = () => {}}) => {
                             })}
                             // eslint-disable-next-line react/jsx-no-bind
                             onClick={() => handleTopicSelect(index)}
+                            role="button"
```
KManolov3: Let's make this a button as well

packages/scratch-gui/src/components/debug-modal/debug-modal.css
kbangelov: I forgot to make a comment about this, I intended to. I think it would be a good idea to have this as a global class in the project, since buttons have default stylings that we so often don't want. Do we have anything of the sort currently and do you think it'd be a good idea to add it? What do you think? cc: @adzhindzhi

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
kbangelov: I tested it and it seems to only call the **handleKeyDownSlides** method once per click. I think what happens is that since **isOpen** is a dependency, it reruns on change, so the cleanup happens even then.

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
```diff
@@ -143,22 +188,32 @@ const DebugModal = ({isOpen, onClose = () => {}}) => {
...
-                            src={nextIcon}
-                            alt="Next"
+                            className={classNames(styles.buttonStyleRemover,
+                                styles.previousIcon,
+                                {[styles.hidden]: selectedTopicIndex === 0})}
+                            onKeyDown={handleKeyDownPrevious}
+                        >
+                            <img
+                                src={prevIcon}
+                                alt="Previous"
```
adzhindzhi: nitpick: Do we want to translate the `alt` texts?

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
```diff
@@ -64,6 +65,48 @@ const DebugModal = ({isOpen, onClose = () => {}}) => {
...
+        if (e.key === KEY.SPACE) {
+            handlePrevious();
+        }
+    }, [handlePrevious]);
+
+    const handleKeyDownNext = useCallback(e => {
+        if (e.key === KEY.SPACE) {
+            handleNext();
+        }
+    }, [handleNext]);
```
adzhindzhi: I was initially a bit torn whether  handling `Space` specifically for activating the buttons makes sense in this context (rather than just leaving the default button logic), but I think I'm fine leaving it as it.

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
adzhindzhi: I think generally React calls the cleanup function whenever the dependency array changes, so we wouldn't end up with multiple event listeners.

adzhindzhi (approved):
I left a few small comments, but overall looks good to me!  🙌

packages/scratch-gui/src/components/debug-modal/debug-modal.jsx
kbangelov: It dawned on me that we don't even need the custom handling methods since button handles enter on its own.

KManolov3 (approved):
Looks good to me - functionally it feels quite intuitive. Good job! 🙌

PR #452, opened 2026-02-20
[UEPR-483] Action Menu Tests

kbangelov:
### Resolves

Part of https://scratchfoundation.atlassian.net/browse/UEPR-483

### Proposed Changes

3 Unit tests to test keyboard navigation.

### Reason for Changes

Part of Accessibility initiative for Scratch.


packages/scratch-gui/test/unit/components/action-menu.test.jsx
```diff
@@ -0,0 +1,118 @@
...
+        const firstItem = screen.getByRole('button', {name: 'Button 1'});
+        expect(document.activeElement).toBe(firstItem);
+
+        act(() => {
+            fireEvent.keyDown(firstItem, {key: KEY.ESCAPE});
+        });
+        expect(document.activeElement).toBe(mainButton);
+    });
+
+    test('tab closes menu and focuses next element', async () => {
```
KManolov3: Maybe we can add a test for shift + tab as well

packages/scratch-gui/test/unit/components/action-menu.test.jsx
```diff
@@ -0,0 +1,118 @@
...
+        const firstItem = screen.getByRole('button', {name: 'Button 1'});
+        expect(document.activeElement).toBe(firstItem);
+
+        await act(async () => {
+            await user.tab();
+        });
+
+        // Wait 1 second for any menu close animations or timeouts
+        await new Promise(resolve => setTimeout(resolve, 1000));
+        expect(document.activeElement).toBe(afterButton);
```
KManolov3: Would we also want to check whether the menu is closed as well?

packages/scratch-gui/test/unit/components/action-menu.test.jsx
```diff
@@ -0,0 +1,118 @@
...
+            fireEvent.keyDown(mainButton, {key: KEY.ARROW_DOWN});
+        });
+        
+        const firstItem = screen.getByRole('button', {name: 'Button 1'});
+        expect(document.activeElement).toBe(firstItem);
+
+        act(() => {
+            fireEvent.keyDown(firstItem, {key: KEY.ESCAPE});
+        });
+        expect(document.activeElement).toBe(mainButton);
```
KManolov3: Here as well - the test name implies we check that the menu is closed, but we only check whether the focus has moved

packages/scratch-gui/test/unit/components/action-menu.test.jsx
kbangelov: Though I did add one, it kept failing if I entered the menu before shift + tab-ing and I couldn't figure out why. Perhaps it's some difference between the user testing library and the real thing. Not sure if it's ok to leave it like this.

KManolov3: I see that you ended up adding a test for that. Did it start working consistently?

kbangelov: The one I left, yes. If we add **arrow_down** to get inside the menu it doesn't perform correctly. Since there were some changes to the base branch since then, I will wait for that one to get merged first and then try again with this test (and check the other ones too).

packages/scratch-gui/test/unit/components/action-menu.test.jsx
```diff
@@ -107,12 +96,41 @@ describe('ActionMenu keyboard navigation', () => {
...
+        const user = userEvent.setup();
 
-        jest.useFakeTimers();
+        act(() => {
+            mainButton.focus();
+        });
+
+        const menuContainer = mainButton.parentElement;
+        expect(menuContainer).toHaveClass('expanded');
+        await act(async () => {
```
KManolov3: Could we use waitFor() instead of having an act with a timeout?

packages/scratch-gui/test/unit/components/action-menu.test.jsx
```diff
@@ -0,0 +1,135 @@
...
+        mockMoreButtonClick.mockClear();
+    });
+
+    test('focus + arrow_down opens menu and arrow_up cycles to last', () => {
+        render(<ActionMenu {...defaultProps} />);
+        const mainButton = screen.getByRole('button', {name: 'Main Button'});
+
+        act(() => {
+            mainButton.focus();
+            fireEvent.keyDown(mainButton, {key: KEY.ARROW_DOWN});
```
adzhindzhi: nitpick: I think the this test case in the way that it's described and written is a bit misleading. Maybe we could do something like:
- Focus opens menu and select the last element by default
- Arrow down on last element cycles to the first
- Arrow up on the first element cycles to the last

packages/scratch-gui/test/unit/components/action-menu.test.jsx
```diff
@@ -0,0 +1,135 @@
...
+        expect(menuContainer).toHaveClass('expanded');
+    });
+
+    test('escape closes menu and returns focus to main button', () => {
+        render(<ActionMenu {...defaultProps} />);
+        const mainButton = screen.getByRole('button', {name: 'Main Button'});
+
+        act(() => {
+            mainButton.focus();
+            fireEvent.keyDown(mainButton, {key: KEY.ARROW_DOWN});
```
adzhindzhi: Why do we need `ARROW_DOWN` to open the menu? Is it just so that we'd go to the first element? I wonder if that's relevant to the current test case, maybe it'd make more sense to keep it simple and avoid doing extra navigations? Same question for the other test cases where we do it.

packages/scratch-gui/test/unit/components/action-menu.test.jsx
```diff
@@ -0,0 +1,135 @@
...
+        });
+
+        const firstItem = screen.getByRole('button', {name: 'Button 1'});
+        expect(document.activeElement).toBe(firstItem);
+
+        act(() => {
+            user.tab();
+        });
```
adzhindzhi: Out of curiosity, what's the difference between calling `tab()` on a user event and firing `KEY.TAB`?

packages/scratch-gui/test/unit/components/action-menu.test.jsx
```diff
@@ -0,0 +1,135 @@
...
+            expect(mainButton.parentElement).not.toHaveClass('expanded');
+        });
+    });
+
+    test('shift + tab closes menu and focuses previous element', async () => {
+        render(
+            <>
+                <button>Before Menu</button>
+                <ActionMenu {...defaultProps} />
+                <button>After Menu</button>
```
adzhindzhi: We probably don't need the after menu in this layout, since we never use it?

packages/scratch-gui/test/unit/components/action-menu.test.jsx
kbangelov: **fireEvent.keyDown** practically seemed to apply some **.preventDefault()** behavior.

packages/scratch-gui/test/unit/components/action-menu.test.jsx
```diff
@@ -30,49 +30,77 @@ describe('ActionMenu keyboard navigation', () => {
...
+            fireEvent.keyDown(mainButton, {key: KEY.ARROW_UP});
+        });
+
+        await waitFor(() => {
+            expect(document.activeElement).toBe(screen.getByRole('button', {name: 'Button 3'}));
+        });
     });
 
-    test('escape closes menu and returns focus to main button', () => {
+    test('arrow_up from first item cycles to last, arrow_down from last cycles to first', async () => {
```
KManolov3: Nitpick: naming -  `ActionMenu keyboard navigation ` + `arrow_up from first item cycle to last, ...` can become `ActionMenu keyboard navigation ` + `cycles to last when arrow_up pressed from first item, ...`

KManolov3: Nitpick: also ideally those would be separate testcases as they don't require that much setup/teardown to justify them being combined, but don't mind it staying that way

packages/scratch-gui/test/unit/components/action-menu.test.jsx
```diff
@@ -82,25 +110,19 @@ describe('ActionMenu keyboard navigation', () => {
...
-
-        act(() => {
-            user.tab();
-        });
+        await user.tab();
 
         await waitFor(() => {
             expect(document.activeElement).toBe(afterButton);
-            expect(mainButton.parentElement).not.toHaveClass('expanded');
+            expect(screen.getByRole('button', {name: 'Main Button'}).parentElement).not.toHaveClass('expanded');
```
KManolov3, line 138: Hasn't this been renamed to `Button 1` now?

KManolov3 (approved):
Minor comments, but overall looks good 🙌

packages/scratch-gui/test/unit/components/action-menu.test.jsx
kbangelov: They are different. Main button is the core one, Button 1 is the top one from the expanded menu.

KManolov3: Ah, yes, fair enough!

PR #454, opened 2026-02-24
[UEPR-517] Return untranslated string when throttling limit is reached for Translate block

kbangelov:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-517

### Proposed Changes

Changes fallback when limit is reached to be untranslated string

### Reason for Changes

UX update to prevent bad fallback when throttling limit is reached.


KManolov3 (approved):
Looks good!

PR #461, opened 2026-03-05
feat: add webp support

mikebarkmin:
### Proposed Changes

Add support for uploading WebP images.

### Reason for Changes

Many images on the web are provided in the WebP format. Because Scratch currently does not support WebP, students often struggle to upload images they find online into their projects. They first need to convert these files into formats like PNG or JPG, which adds unnecessary complexity and interrupts the creative workflow.

Adding WebP support would make it much easier for students to use images from the web directly in Scratch. This would reduce frustration, save time, and allow learners to focus more on creativity and programming rather than dealing with file format limitations.

### Test Coverage

I added a test similar to uploading an bmp image here: https://github.com/mikebarkmin/scratch-editor/blob/4f08230c5ef1a2d404f106e72e415a8c40dc715c/packages/scratch-gui/test/integration/costumes.test.js#L164


mikebarkmin:

mikebarkmin:
recheck

PR #470, opened 2026-03-12
fix: fix de/serialization for non-shadow proc defs

### LINKED ISSUE #3472: Procedure definitions don't properly survive save/load
## Expected Behavior

Saving and reloading a project should result in the same arrangement of blocks that were present before the save.

## Actual Behavior

Procedure definitions can break in a variety of ways. The simplest is that the argument reporter(s) in the prototype block can get duplicated, leaving a non-deletable, disconnected argument reporter stuck on the workspace. In more complicated projects, like ["Eat Pi!"](https://scratch.mit.edu/projects/1286790763/editor/), it can cause major problems in block deserialization.

## Steps to Reproduce

Simplest case:

1. Use "Make a Block" to create a custom procedure with at least one argument
2. Use File -> "Save to your computer" to save an SB3 file
3. Use File -> "Load from your computer" to load that SB3 file
4. Look at the weird, broken block(s)

More complicated case: replace steps 1 & 2 with opening a complex project from our main website, saving it to an SB3 file. Then, continue with step 3 on a Sporked build.

cwillisf:
### Resolves

- Resolves scratchfoundation/scratch-blocks#3472

### Proposed Changes

Update block de/serialization such that procedure definitions and their argument reporters are no longer saved as shadow blocks, and cope with previously-saved projects which already had them serialized as shadow blocks.

### Reason for Changes

When loading from SB3, the XML sent to Blockly included argument reporters as children of `procedures_prototype`. Blockly's `domToMutation` created a new set before those children connected, displacing the mutation-created reporters and leaving them floating on the workspace as undeletable phantoms. Fix by stripping arg reporter children from prototype blocks during deserialization, and correct legacy shadow flags on prototypes for both SB2 and SB3 formats.

Related to scratchfoundation/scratch-blocks#3465

### Test Coverage

Tested manually; see repro notes in scratchfoundation/scratch-blocks#3465

### Thanks

Big thanks to @KManolov3 for bringing this issue to my attention

### DIFF
--- a/packages/scratch-vm/src/serialization/sb2.js
+++ b/packages/scratch-vm/src/serialization/sb2.js
@@ -1240,7 +1240,7 @@ const parseBlock = function (sb2block, addBroadcastMsg, getVariableId, extension
         activeBlock.inputs[inputName] = {
             name: inputName,
             block: inputUid,
-            shadow: inputUid
+            shadow: null
         };
         activeBlock.children = [{
             id: inputUid,
@@ -1248,7 +1248,7 @@ const parseBlock = function (sb2block, addBroadcastMsg, getVariableId, extension
             inputs: {},
             fields: {},
             next: null,
-            shadow: true,
+            shadow: false,
             children: [],
             mutation: {
                 tagName: 'mutation',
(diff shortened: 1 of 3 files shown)


KManolov3 (approved):
Not acquainted well enough to understand everything that's going here, but generally looks good to me!

PR #471, opened 2026-03-13
[UEPR-518] Implement in-editor Manual Save Project Thumbnail logic

kbangelov:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-518

### Proposed Changes

Moves the "update thumbnail" button from project page to project editor and adds:
- a modal to confirm action
- alerts for updating status, confirmation and fail message
- blue loading circle

Design here : https://www.figma.com/design/nQK2PbAYG7pCmK4lYOTvHZ/Save-Thumbnail?node-id=370-1473&p=f&m=dev

### To Do
- show feature tooltip for moved button on first open of editor per user ONLY.


packages/scratch-gui/src/components/alerts/alert.css
```diff
@@ -5,6 +5,7 @@
 body .alert {
     width: 100%;
     display: flex;
+    box-sizing: border-box;
```
kbangelov, line 8: This is a generic change that applies to other alerts too. But is there any point in that extra space on the right otherwise?

kbangelov (commented):
I have not changed delete confirmation prompt to use our new element since it has some visual differences and such a change has not been communicated with the Scratch team yet

packages/scratch-gui/src/lib/alerts/index.jsx
```diff
@@ -12,6 +13,7 @@ const AlertTypes = keyMirror({
 
 const AlertLevels = {
     SUCCESS: 'success',
+    BLUE: 'blue',
```
kbangelov: Perhaps it's not the most descriptive name. Suggestions are welcome.

packages/scratch-gui/src/components/alerts/alert.css
KManolov3: Can you post screenshots to compare before/after?

packages/scratch-gui/src/components/alerts/alert.css
```diff
@@ -30,6 +31,11 @@ body .alert {
     box-shadow: 0px 0px 0px 2px $extensions-light;
 }
 
+.alert.blue {
+    border: 1px solid #4C97FF;
```
KManolov3: Let's extract colors in "colors.css" and use them from there

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.css
```diff
@@ -0,0 +1,70 @@
+@import "../../css/colors.css";
+@import "../../css/units.css";
+
+.modal-container {
+    display: flex;
+    width: 200px;
```
KManolov3: Let's use units relative to the user's font size preferences (rem)

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.css
```diff
@@ -0,0 +1,70 @@
...
+    align-items: flex-start;
+    border-radius: 8px;
+    background: #855CD6;
+    gap: 15px;
+    box-sizing: border-box;
+}
+
+.label {
+    align-self: stretch;
+    color: var(--White-White, #FFF);
```
KManolov3: Does the var (`--White-White`) actually exist? Let's define the color and use it here - we don't need the fallback

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.css
```diff
@@ -0,0 +1,70 @@
...
+
+.button-row {
+    font-weight: bolder;
+    display: flex;
+    gap: 0.5rem;
+    width: 100%;
+}
+
+.button-row button {
+    all: unset;
```
KManolov3, line 34: Maybe we can override the properties we need changed instead of unsetting everything?

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.css
```diff
@@ -0,0 +1,70 @@
...
+    display: flex;
+    padding: 8px 16px;
+    justify-content: center;
+    align-items: center;
+    gap: 8px;
+    flex: 1 0 0;
```
KManolov3: So just `flex: 1`?

packages/scratch-gui/src/components/confirmation-prompt/icon--arrow-up.svg
KManolov3, line 1: Why is this one bigger than the others? What method did you use for exporting them from Figma?

packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -545,6 +540,10 @@ const GUIComponent = props => {
                                 vm={vm}
                                 ariaRole="region"
                                 ariaLabel={intl.formatMessage(ariaMessages.stage)}
+                                manuallySaveThumbnails={manuallySaveThumbnails}
+                                userOwnsProject={userOwnsProject}
+                                onUpdateProjectThumbnail={onUpdateProjectThumbnail}
+                                isInEditor
```
KManolov3: Aren't we passing `onUpdateProjectThumbnail` only here? So is there a case where we wouldn't be in the editor?

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -165,17 +209,33 @@ const StageHeaderComponent = function (props) {
                 <Box className={styles.stageMenuWrapper}>
                     <Controls vm={vm} />
                     <div className={styles.stageSizeRow}>
+                        {manuallySaveThumbnails && isInEditor && isProjectLoaded && userOwnsProject && (
+                            <Button
+                                aria-label={intl.formatMessage(messages.setThumbnail)}
+                                title={intl.formatMessage(messages.setThumbnail)}
+                                className={styles.stageButton}
+                                onClick={onThumbnailPromptOpen}
+                                // disabled={isUpdatingThumbnail}
```
KManolov3: Reminder to update this before merging

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -69,28 +85,56 @@ const StageHeaderComponent = function (props) {
...
+        userOwnsProject,
+        showThumbnailSetting,
+        showThumbnailSuccess,
+        showThumbnailError
     } = props;
     const intl = useIntl();
 
     let header = null;
 
+    const [isThumbnailPromptOpen, setIsThumbnailPromptOpen] = React.useState(false);
```
KManolov3: Nitpick: we can import useState

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -69,28 +85,56 @@ const StageHeaderComponent = function (props) {
...
-                    return;
-                }
+                if (!onUpdateProjectThumbnail) return;
+
+                showThumbnailSetting();
 
                 storeProjectThumbnail(vm, dataURI => {
-                    onUpdateProjectThumbnail(projectId, dataURItoBlob(dataURI));
+                    try {
+                        onUpdateProjectThumbnail(projectId, dataURItoBlob(dataURI)).then();
```
KManolov3: .then() ?

KManolov3: Let's make the callback passed to `storeProjectThumbnail` async and await `onUpdateProjectThumbnail`, so that the error handling works.

If there is an async method in a try catch which is not awaited, the error will not buble up to the catch block

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -69,28 +85,56 @@ const StageHeaderComponent = function (props) {
...
+                        showThumbnailSuccess();
+                    } catch (e) {
+                        showThumbnailError();
+                    }
                 });
             },
             3000
         ),
-        [projectId, onUpdateProjectThumbnail]
+        [onUpdateProjectThumbnail, projectId, showThumbnailSetting, showThumbnailSuccess, showThumbnailError, vm]
```
KManolov3: The vm is passed as an argument to the callback of storeProjectThumbnail, it doesnt need to be a dependency here

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -165,17 +209,33 @@ const StageHeaderComponent = function (props) {
                 <Box className={styles.stageMenuWrapper}>
                     <Controls vm={vm} />
                     <div className={styles.stageSizeRow}>
+                        {manuallySaveThumbnails && isInEditor && isProjectLoaded && userOwnsProject && (
+                            <Button
+                                aria-label={intl.formatMessage(messages.setThumbnail)}
```
KManolov3: It might be better to omit this aria-label, as it doesn't bring any further information - it's the same as the title, which will be read by screen readers

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -198,10 +258,22 @@ const StageHeaderComponent = function (props) {
...
-    stageSizeMode: state.scratchGui.stageSize.stageSize
+const mapStateToProps = state => {
+    const projectState = state.scratchGui.projectState;
+    const loadingState = projectState.loadingState;
+
+    return {
+        projectId: projectState.projectId,
+        // This is the button's mode, as opposed to the actual current state
+        stageSizeMode: state.scratchGui.stageSize.stageSize,
+        isProjectLoaded: getIsShowingWithId(loadingState) || getIsUpdating(loadingState)
```
KManolov3: Are we sure this amounts to the project being loaded?

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -198,10 +258,22 @@ const StageHeaderComponent = function (props) {
...
+    };
+};
+
+const mapDispatchToProps = dispatch => ({
+    showThumbnailSetting: () => dispatch(showStandardAlert('settingThumbnail')),
```
KManolov3: Can we either:
- extract the alertIds in an enum, or
- add a task for doing so in the future?

KManolov3: Nitpick: maybe it will be clearer if those were names such as:
`onShowSettingThumbnail`
`onShowThumbnailSuccess`
`onShowThumbnailError`

packages/scratch-gui/src/lib/alerts/index.jsx
```diff
@@ -229,6 +231,52 @@ const alerts = [
...
+        content: (
+            <FormattedMessage
+                defaultMessage="Thumbnail set successfully."
+                description="Message indicating that the thumbnail was successfully set"
+                id="gui.alerts.thumbnailSuccess"
+            />
+        ),
+        iconURL: successImage,
+        level: AlertLevels.SUCCESS,
+        maxDisplaySecs: 3,
```
KManolov3: How did you decide on 3 seconds? I don't necessarily mind it just curious - I see 5 is the most common value for the maxDisplaySecs

packages/scratch-gui/src/lib/alerts/index.jsx
KManolov3: Maybe `loading`, with a comment that older instances of the alert prefer using `info` AlertLevel ? Or just `info_blue` 😄

packages/scratch-gui/src/components/stage-header/stage-header.css
```diff
@@ -46,14 +47,6 @@
     border-radius: $form-radius;
 }
 
-[dir="ltr"] .stage-size-toggle-group {
```
KManolov3, line 49: Aren't those still used?

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
```diff
@@ -0,0 +1,256 @@
...
+    }
+});
+
+const modalWidth = 200;
+const spaceForArrow = 16;
+const arrowOffsetFromEnd = 7;
+const arrowLongSide = 29;
+const arrowShortSide = 13;
+
+const calculateModalPosition = (relativeElemRef, modalRef, modalPosition) => {
```
KManolov3: I wonder if this can be simplified

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.css
KManolov3: Oh, I see this was taken directly from Figma - we don't want/need to add all the css values from Figma, but rather adapt our code to achieve the same CSS. For example, `--White-White` exists as a var in Figma but not in our code. Additionally, Helvetica Neue is the family that should be set by default in our code.

As a rule of thumb, let's try to simplify the css where some styles aren't necessary, so we don't add any noise.

KManolov3 (commented):
Good job! Overall, comments are minor, but I do wonder if the logic in confirmation-prompt can be simplified.

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -165,17 +241,58 @@ const StageHeaderComponent = function (props) {
...
+                                    src={thumbnailIcon}
+                                    alt={intl.formatMessage(messages.setThumbnail)}
+                                    className={styles.stageButtonIcon}
+                                />
+                            </Button>
+                        )}
+                        <ConfirmationPrompt
+                            isOpen={isThumbnailPromptOpen}
+                            title={messages.setThumbnail}
+                            message={<FormattedMessage {...messages.setThumbnailMessage} />}
```
kbangelov: The props being passed seem inconsistent here - strings in one place and FormattedMessage for the other props. I wonder if all strings makes more sense here?

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -165,17 +241,58 @@ const StageHeaderComponent = function (props) {
...
+                            isOpen={isThumbnailTooltipOpen}
+                            onRequestOpen={onOpenTooltip}
+                            onRequestClose={onCloseTooltip}
+                            targetRef={thumbnailButtonRef}
+                            primaryPosition="left"
+                            secondaryPosition="down"
+                            width={336}
+                            title={intl.formatMessage(messages.thumbnailTooltipTitle)}
+                            body={
+                                <FormattedMessage
```
kbangelov, line 243: I think I need to pass the node here because I don't see how else I'll be able to bold the text just inside the Tooltip component.

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.css
adzhindzhi: Plus one to this, Figma usually adds additional styles that we often don't need. Some of the styles are already set on a more global level, others don't really make a difference for our specific layout so it's usually better to clean them up and simplify the css as much as we can to make it easier for reviewers and future maintainers of the code. 

Also in general, wе should try not to use hardcoded colors, but always check if it's defined in colors.css first, and if not - add it.

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
```diff
@@ -0,0 +1,190 @@
...
+        description: 'Label for cancel button in confirmation prompt',
+        id: 'gui.confirmationPrompt.cancel'
+    }
+});
+
+const modalWidth = 200;
+const spaceForArrow = 16;
+const arrowOffsetFromEnd = 7;
+const arrowLongSide = 29;
+const arrowShortSide = 13;
```
adzhindzhi: If we truly want to make this component generic, we should make these configurable.

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
```diff
@@ -0,0 +1,190 @@
...
+
+const ConfirmationPrompt = ({
```
adzhindzhi, line 58: Can we reuse this new ConfirmationPrompt for the `DeleteConfirmationPrompt`?

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -165,17 +241,58 @@ const StageHeaderComponent = function (props) {
...
+                        <Tooltip
+                            isOpen={isThumbnailTooltipOpen}
+                            onRequestOpen={onOpenTooltip}
+                            onRequestClose={onCloseTooltip}
+                            targetRef={thumbnailButtonRef}
+                            primaryPosition="left"
+                            secondaryPosition="down"
+                            width={336}
+                            title={intl.formatMessage(messages.thumbnailTooltipTitle)}
+                            body={
```
adzhindzhi, line 242: The tooltip display is generally a part of another task - https://scratchfoundation.atlassian.net/browse/UEPR-522, so we don't need to handle it here. There are also more specific requirements that we need to implement, but I'd suggest leaving this as is for now and then finishing/updating the callout work as part of the other task.

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -165,17 +241,58 @@ const StageHeaderComponent = function (props) {
...
+                        )}
+                        <ConfirmationPrompt
+                            isOpen={isThumbnailPromptOpen}
+                            title={messages.setThumbnail}
+                            message={intl.formatMessage(messages.setThumbnailMessage)}
+                            onConfirm={onUpdateThumbnailAndClose}
+                            onCancel={onThumbnailPromptClose}
+                            relativeElementRef={thumbnailButtonRef}
+                            primaryPosition="down"
+                            secondaryPosition="left"
```
adzhindzhi: Maybe rename these to `side` and `align`? I think it'd be easier for readers to understand which is which. 
`side` - On which side of the anchor element should the prompt be displayed (top, bottom, left, right)
`align` - Where should the arrow be placed (left, center, right)

If we go with this approach, we'd also have to flip the logic for `left` and `right` for the secondary position. Right now, `left` indicates that most of the content is displayed on the left side of the button, but the arrow is displayed on the right side of the modal itself, which was not what I expected seeing this configuration.

packages/scratch-gui/src/hooks/calculatePopupPosition.js
```diff
@@ -0,0 +1,99 @@
...
+    arrowUpIcon,
+    arrowDownIcon,
+    spaceForArrow,
+    arrowShortSide,
+    arrowLongSide,
+    arrowOffsetFromEnd,
+    arrowOffsetFromBottom = 0
+}) => {
+    const modalHeight = popupRef.current.getBoundingClientRect().height;
+    const arrowHeight = (primaryPosition === 'left' || primaryPosition === 'right') ? arrowLongSide : arrowShortSide;
```
KManolov3: Nitpick - maybe we could
````
const [arrowHeight, arrowWidth] = (primaryPosition === 'left' || primaryPosition === 'right') ? [arrowLongSide, arrowShortSide] : [arrowShortSide, arrowLongSide];
````

File: packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx (hunk 22 above)
Comment (KManolov3) [inline packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx:58]: Can we:
- pass the positioning params (width, arrow sizes, ..) as a configuration object to the ConfirmationPrompt
- reuse this component for the delete confirmation prompt, since it already contains a less complex version of the logic used here

File: packages/scratch-gui/src/hooks/calculatePopupPosition.js [hunk 26]
```diff
@@ -0,0 +1,99 @@
...
+    case 'down':
+        top = buttonRect.bottom + spaceForArrow;
+        break;
+    case 'left':
+        left = buttonRect.left - popupWidth - spaceForArrow;
+        break;
+    case 'right':
+        left = buttonRect.right + spaceForArrow;
+        break;
+    }
```
adzhindzhi: Let's add these values to an enum, instead of hardcoding them

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
```diff
@@ -0,0 +1,190 @@
...
+            const pos = useCalculatePopupPosition({
+                relativeElementRef,
+                popupRef: modalRef,
+                primaryPosition,
+                secondaryPosition,
+                popupWidth: modalWidth,
+                arrowLeftIcon,
+                arrowRightIcon,
+                arrowUpIcon,
+                arrowDownIcon,
```
adzhindzhi: Why do we add the icon as part of the position? Is it simply for checking `primaryPosition`? We could probably have a map defined here for `primaryPosition` (or `side`) -> `arrowIcon ` and use that directly, instead of as part of the position. It'd also reduce the number of parameters we pass to the hook.

packages/scratch-gui/src/lib/calculatePopupPosition.js
```diff
@@ -0,0 +1,99 @@
...
+    popupWidth,
+    arrowLeftIcon,
+    arrowRightIcon,
+    arrowUpIcon,
+    arrowDownIcon,
+    spaceForArrow,
+    arrowShortSide,
+    arrowLongSide,
+    arrowOffsetFromEnd,
+    arrowOffsetFromBottom = 0
```
adzhindzhi: Hmmm, I wonder if we can avoid any of these by simply passing `arrowOffset`? For example, what's the difference between `spaceForArrow` and `arrowOffsetFromBottom`? I think we simply need to how the `arrowHeight` and the `offset` from the anchor element? 

Also, `arrowOffsetFromEnd` is a bit specific, I wonder if this should be something configurable at all. If anything, it should at least have a default value, because the cases of when we'd want to change that would be rare.

packages/scratch-gui/src/components/stage-header/stage-header.jsx
KManolov3: I can also see `arrowPosition` as an alternative naming for the `align` prop

packages/scratch-gui/src/lib/calculatePopupPosition.js
KManolov3: II agree, let's add enough configuration for the cases we currently have, and if needed we can always expand the configuration later.

packages/scratch-gui/src/components/alerts/alert.css
kbangelov: Before:
Screenshot from 2026-03-18 15-39-20

After:
Screenshot from 2026-03-18 15-35-14

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.css
kbangelov: I think it's practically all except **cursor:pointer;**

packages/scratch-gui/src/components/gui/gui.jsx
kbangelov: That's true, we don't need both checks.

packages/scratch-gui/src/components/stage-header/stage-header.jsx
kbangelov: some leftover

packages/scratch-gui/src/components/stage-header/stage-header.jsx
kbangelov: It amounts to the project being loaded except for project being loaded without id. I wonder if we should cover that too? I am not too sure when it is reached?

packages/scratch-gui/src/lib/alerts/index.jsx
kbangelov: Tereza agreed on 3. You do have a point though. I see no reason to diverge from the common timeout. It is not as short of a message as "Project saved."

packages/scratch-gui/src/components/stage-header/stage-header.css
kbangelov: The design in the task description doesn't have that extra margin which would make the gaps uneven. Tereza agreed on 4px (.25 rem) for both gaps

packages/scratch-gui/src/components/stage-header/stage-header.jsx
kbangelov: since **secondaryPosition** is more about where the entire modal is placed (instead of centrally) I'll rename it to **align**

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -69,28 +88,88 @@ const StageHeaderComponent = function (props) {
...
+    const [isThumbnailTooltipOpen, setIsThumbnailTooltipOpen] = useState(false);
+    const [isUpdatingThumbnail, setIsUpdatingThumbnail] = useState(false);
+
+    // To remove - new feature awareness tooltip
+    useEffect(() => {
+        if (manuallySaveThumbnails && isProjectLoaded &&
+            userOwnsProject && thumbnailButtonRef.current) {
+            setIsThumbnailTooltipOpen(true);
+        }
+    }, [manuallySaveThumbnails, isProjectLoaded, userOwnsProject]);
```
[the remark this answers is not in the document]
kbangelov: Not to be handled in this PR

packages/scratch-gui/src/components/tooltip/tooltip.css
```diff
@@ -0,0 +1,42 @@
+@import "../../css/colors.css";
+
+.tooltip {
+    display: flex;
+    flex-direction: column;
+    width: 21rem;
+    padding: 1rem;
+    align-items: flex-start;
+    gap: -0.0625rem;
```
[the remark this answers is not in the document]
kbangelov: not my change but perhaps you're right. I just made it rem instead of px

packages/scratch-gui/src/components/confirmation-prompt/icon--arrow-up.svg
kbangelov: There was some gap on the originally downloaded one. I wonder if it's worth resolving, since it has the same proportions and same length in the end.

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.css
```diff
@@ -0,0 +1,70 @@
...
+    background: $looks-secondary;
+    gap: 0.9375rem;
+    box-sizing: border-box;
+}
+
+.label {
+    align-self: stretch;
+    color: $ui-white;
+    text-align: center;
+    font-family: "Helvetica Neue";
```
KManolov3: This is probably redundant

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.css
```diff
@@ -0,0 +1,70 @@
...
+    font-size: 1rem;
+    font-style: normal;
```
KManolov3: Nitpick: is this needed?

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.css
```diff
@@ -0,0 +1,70 @@
...
+    font-weight: 700;
+    line-height: 1.5rem;
+}
+
+.button-row {
+    font-weight: bolder;
```
KManolov3: Nitpick: would it be clearer to use a numeric value here as well?

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.css
```diff
@@ -0,0 +1,70 @@
...
+    align-items: center;
+    gap: 0.5rem;
+    flex: 1;
+    border-radius: 2.5rem;
+    background: inherit;
+    cursor: pointer;
+}
+
+.button-row button:focus {
+    outline: auto;
```
KManolov3, line 49: Is this needed?

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
```diff
@@ -0,0 +1,218 @@
...
+    message,
+    confirmLabel,
+    cancelLabel,
+    onConfirm,
+    onCancel,
+    isOpen,
+    relativeElementRef,
+    side,
+    align,
+    config
```
KManolov3: Nitpick: can we use a more precise name, such as `layoutConfig`?

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
```diff
@@ -0,0 +1,218 @@
...
+
+    const updatePosition = useCallback(() => {
+        if (relativeElementRef.current && modalRef.current) {
+            const pos = calculatePopupPosition({
+                relativeElementRef,
+                popupRef: modalRef,
+                side,
+                align,
+                popupWidth: modalWidth,
+                arrowLeftIcon,
```
KManolov3: nitpick: maybe we can just pass an arrowIcons object containing the arrows for different sides.

KManolov3: Additionally, if the icons are the same, could we pass a single arrow icon and flip it in the desired direction with css?

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
```diff
@@ -0,0 +1,218 @@
...
+            setModalPositionValues(pos);
+        }
+    }, [
+        relativeElementRef,
+        side,
+        align,
+        modalWidth,
+        spaceForArrow,
+        counterOffset,
+        arrowShortSide,
```
KManolov3: What did `arrowShortSide` / `arrowLongSide` denote?

packages/scratch-gui/src/components/stage-header/icon--thumbnail.svg
KManolov3, line 1: This doesn't look like it visualises in github. Is it as intended to be?

Image

packages/scratch-gui/src/components/spinner/spinner.css
```diff
@@ -29,6 +33,10 @@
     animation: spin 1.5s cubic-bezier(0.4, 0.1, 0.4, 1) infinite;
 }
 
+.spinner.info-blue::after {
+    border-top-color: $motion-primary;
```
KManolov3, line 37: Nitpick: Why do we need this?

packages/scratch-gui/src/components/stage-header/stage-header.css
KManolov3: Even so, I believe they're still being referred in the code

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -69,28 +88,88 @@ const StageHeaderComponent = function (props) {
...
     let header = null;
 
+    const thumbnailTooltipId = 'thumbnail-tooltip';
+    const thumbnailButtonRef = useRef(null);
+
+    const [isThumbnailPromptOpen, setIsThumbnailPromptOpen] = useState(false);
+    const [isThumbnailTooltipOpen, setIsThumbnailTooltipOpen] = useState(false);
+    const [isUpdatingThumbnail, setIsUpdatingThumbnail] = useState(false);
+
+    // To remove - new feature awareness tooltip
```
KManolov3: Nitpick, not insisting in this task: we should create a new task for removing this after 60 days and refer it in this TODO

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -69,28 +88,88 @@ const StageHeaderComponent = function (props) {
...
+
+                setIsUpdatingThumbnail(true);
+                onShowSettingThumbnail();
 
-                storeProjectThumbnail(vm, dataURI => {
-                    onUpdateProjectThumbnail(projectId, dataURItoBlob(dataURI));
-                });
+                try {
+                    await storeProjectThumbnail(vm, dataURI => new Promise((resolve, reject) => {
+                        onUpdateProjectThumbnail(
```
KManolov3: We have two options on how to handle this:
- either update `onUpdateProjectThumbnail` to return a Promise itself, so it can be awated, or
- update `onUpdateProjectThumbnail` to accept and call the resolve/reject callbacks (as implied here)

In either case, we should coordinate with @adzhindzhi, since she's updating www

packages/scratch-gui/src/components/stage-header/stage-header.jsx
KManolov3: You can consider `intl.formatMessage` as well

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -165,17 +244,57 @@ const StageHeaderComponent = function (props) {
...
+                        )}
+                        <ConfirmationPrompt
+                            isOpen={isThumbnailPromptOpen}
+                            title={intl.formatMessage(messages.setThumbnail)}
+                            message={intl.formatMessage(messages.setThumbnailMessage)}
+                            onConfirm={onUpdateThumbnailAndClose}
+                            onCancel={onThumbnailPromptClose}
+                            relativeElementRef={thumbnailButtonRef}
+                            side="down"
+                            align="left"
```
KManolov3: Let's have the possible options for these in enums

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -165,17 +244,57 @@ const StageHeaderComponent = function (props) {
...
+                            <Button
+                                title={intl.formatMessage(messages.setThumbnail)}
+                                className={classNames(
+                                    styles.stageButton,
+                                    {[styles.stageButtonHighlighted]: isThumbnailTooltipOpen}
+                                )}
+                                onClick={onThumbnailPromptOpen}
+                                disabled={isUpdatingThumbnail}
+                                componentRef={thumbnailButtonRef}
+                                data-tip={intl.formatMessage(messages.setThumbnail)}
```
KManolov3: Do you need to pass this, since it's the same as the title?

packages/scratch-gui/src/components/tooltip/tooltip.css
```diff
@@ -0,0 +1,42 @@
...
+
+    border-radius: 0.5rem;
+    border: 0.0625rem solid $motion-primary;
+    background-color: $motion-primary;
+    opacity: 1 !important;
```
KManolov3: Why do we need the important? Please leave as a comment in the code as well

packages/scratch-gui/src/components/tooltip/tooltip.css
KManolov3: Are you sure a negative gap is what you want? Or do you indeed want to reduce paddings / margins?

packages/scratch-gui/src/components/tooltip/tooltip.css
```diff
@@ -0,0 +1,42 @@
...
+
+    /* Basic Shadows/2dp */
+    box-shadow: 0 0.25rem 0.375rem 0 $ui-black-transparent-10, 0 0.25rem 1rem 0 $ui-black-transparent;
+}
+
+.tooltip-title {
+    align-self: stretch;
+    color: $ui-white;
+    font-family: "Helvetica Neue";
```
KManolov3: redundant?

packages/scratch-gui/src/components/tooltip/tooltip.jsx
```diff
@@ -0,0 +1,233 @@
...
+        arrowOffsetFromBottom,
+        arrowLongSide,
+        arrowShortSide
+    } = {...defaultConfig, ...config};
+    const [arrowHeight, arrowWidth] = (side === PopupSide.LEFT || side === PopupSide.RIGHT) ?
+        [arrowLongSide, arrowShortSide] : [arrowShortSide, arrowLongSide];
+
+    const updatePosition = useCallback(() => {
+        if (!targetRef?.current || !tooltipRef.current) return;
+        const newPos = calculatePopupPosition({
```
KManolov3: I see a lot of similarities between the tooltip and the confirmation modal - the calls to `calculatePopupPosition`, the layout config, the arrowHeight / width. Do you see potential in extracting a common component between the two?

packages/scratch-gui/src/components/tooltip/tooltip.jsx
```diff
@@ -0,0 +1,233 @@
...
+        window.addEventListener('resize', updatePosition);
+        window.addEventListener('scroll', updatePosition, true);
+        return () => {
+            window.removeEventListener('resize', updatePosition);
+            window.removeEventListener('scroll', updatePosition, true);
+        };
+    }, [isOpen, updatePosition]);
+
+    // Click outside to close
+    useEffect(() => {
```
KManolov3: Such behaviors seem like we are implementing parts of the standard <Modal behaviour. Is there any potential in using a modal component underneath?

packages/scratch-gui/src/components/tooltip/tooltip.jsx
```diff
@@ -0,0 +1,233 @@
...
+
+    // Simulate hover and focus (normal) tooltip behavior
+    useEffect(() => {
+        if (isManualOnly) return;
+        
+        const target = targetRef?.current;
+        if (!target) return;
+
+        const handleMouseEnter = () => {
+            if (onRequestOpen) onRequestOpen();
```
KManolov3: Nitpick: `if (onRequestOpen) onRequestOpen()` -> `onRequestOpen?.()`

packages/scratch-gui/src/components/tooltip/tooltip.jsx
```diff
@@ -0,0 +1,233 @@
...
+        target.addEventListener('focus', handleFocus);
+        target.addEventListener('blur', handleBlur);
+
+        return () => {
+            target.removeEventListener('mouseenter', handleMouseEnter);
+            target.removeEventListener('mouseleave', handleMouseLeave);
+            target.removeEventListener('focus', handleFocus);
+            target.removeEventListener('blur', handleBlur);
+        };
+    }, [isManualOnly, onRequestOpen, onRequestClose, targetRef, targetRef?.current]);
```
KManolov3: Nitpick: isn't `targetRef, targetRef?.current` equivalent to just `targetRef?.current` ?

packages/scratch-gui/src/lib/store-project-thumbnail.js
```diff
@@ -1,20 +1,24 @@
...
     } catch (e) {
         log.error('Project thumbnail save error', e);
-        // This is intentionally fire/forget because a failure
-        // to save the thumbnail is not vitally important to the user.
+        throw e; // re-throw so it is handled by alert
     }
 };
 
-export const getProjectThumbnail = (vm, callback) => {
+export const getProjectThumbnail = (vm, callback) => new Promise((resolve, reject) => {
```
KManolov3: If we are going to promisify `getProjectThumbnail`, then we have no need to pass the callback to it. We can make it resolve with the `dataURI`, and then the caller can decide what to do with it (in our case `storeProjectThumbnail` wants to invoke its callback on it.

packages/scratch-gui/src/lib/store-project-thumbnail.js
```diff
@@ -1,20 +1,24 @@
...
     vm.postIOData('video', {forceTransparentPreview: true});
-    vm.renderer.requestSnapshot(dataURI => {
+    vm.renderer.requestSnapshot(async dataURI => {
         vm.postIOData('video', {forceTransparentPreview: false});
-        callback(dataURI);
+        try {
+            await callback(dataURI);
+            resolve();
```
KManolov3: I wonder whether we'd want to invoke `vm.renderer.draw()` before resolving the promise - otherwise we won't be notified about an error there (which might still lead to crashing the editor)

packages/scratch-gui/src/lib/store-project-thumbnail.js
```diff
@@ -1,20 +1,24 @@
...
+        } catch (e) {
+            reject(e instanceof Error ? e : new Error(String(e)));
```
KManolov3: Nitpick: imo just reject with e here

packages/scratch-gui/src/lib/store-project-thumbnail.js
```diff
@@ -1,20 +1,24 @@
...
+export const storeProjectThumbnail = async (vm, callback) => {
     try {
-        getProjectThumbnail(vm, callback);
+        await getProjectThumbnail(vm, callback);
     } catch (e) {
         log.error('Project thumbnail save error', e);
-        // This is intentionally fire/forget because a failure
-        // to save the thumbnail is not vitally important to the user.
+        throw e; // re-throw so it is handled by alert
     }
```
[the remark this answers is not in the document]
KManolov3: I agree - if we are updating this to be async, we should update all usages (so including the `project-saver-hoc`

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.css
kbangelov: currently yes because of the earlier **all: unset;** on the button

packages/scratch-gui/src/components/stage-header/icon--thumbnail.svg
kbangelov: oh no haha that is the icon for the thumbnail, so it does visualise

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
kbangelov: Height and width depending on direction. I renamed them for clarity.

packages/scratch-gui/src/components/stage-header/icon--thumbnail.svg
KManolov3: Fair enough 😅

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
kbangelov: I tried it out and turned out to be a needlessly hard geometry exercise. I'll keep all of the SVGs for now. For the comment above, I don't even need to pass them to the calculate... method.

packages/scratch-gui/src/components/spinner/spinner.css
kbangelov: To make the thing that moves in the spinner circle blue, as in the design.

packages/scratch-gui/src/components/stage-header/stage-header.css
kbangelov: That was only needed previously when it was considered the leftmost and only element in that group. Now I add the gap from the parent component, so I removed the div entirely.

packages/scratch-gui/src/components/tooltip/tooltip.css
kbangelov: It is that way by design. I guess it's not terribly important whether we have it or not but I don't see a reason to diverge.

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.css
adzhindzhi: I wonder if that's a sign not to use `all:unset` but rather change what actually needs to be changed? My concern is that we may unset some default button styles that we haven't though of testing. I don't have a strong preference though, that's just a thought.

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.css
```diff
@@ -0,0 +1,68 @@
...
+}
+
+.button-row button span {
+    color: $ui-white;
+    font-family: "Helvetica Neue";
```
adzhindzhi: Do we need to specify the `font-family` everywhere - e.g. here, in `.label`? Isn't that a global style?

packages/scratch-gui/src/components/stage-header/stage-header.css
```diff
@@ -75,26 +68,19 @@
     height: 100%;
 }
 
+.stage-button-highlighted {
+    border-radius: 4px;
+    box-shadow: 0 0 0 2px #4C97FF;
```
adzhindzhi: Let's move use this color from `colors.css`

packages/scratch-gui/src/components/stage-header/stage-header.jsx
adzhindzhi: Hmm, `onUpdateProjectThumbnail` already accepts `onSuccess`, `onError`. Can we not pass the callbacks directly? Why do we need to await the promises?

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -139,43 +218,79 @@ const StageHeaderComponent = function (props) {
...
+                            targetRef={thumbnailButtonRef}
+                            side={PopupSide.LEFT}
+                            align={PopupAlign.DOWN}
+                            layoutConfig={{width: 336}}
+                            title={intl.formatMessage(messages.thumbnailTooltipTitle)}
+                            body={
+                                <FormattedMessage
+                                    {...messages.thumbnailTooltipBody}
+                                    values={{
+                                        boldText: <b>{intl.formatMessage(messages.setThumbnail)}</b>
```
adzhindzhi: Instead of using `formatMessage` inside a `FormattedMessage`, you can do something like what I've done in [this PR](https://github.com/scratchfoundation/scratch-www/pull/10113/changes#diff-8bb8f860d4bc076d0247afbc29a9419ba507d1303416d865d02d46ce947e205fR77-R88). Basically that way you can simply add `<b>...</b>` in the message definition and that would bold it.

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -198,12 +313,6 @@ const StageHeaderComponent = function (props) {
     return header;
 };
 
-const mapStateToProps = state => ({
-    projectId: state.scratchGui.projectState.projectId,
-    // This is the button's mode, as opposed to the actual current state
-    stageSizeMode: state.scratchGui.stageSize.stageSize
-});
-
```
adzhindzhi, line 206: Don't we still need this for the stage size mode?

packages/scratch-gui/src/components/confirmation-prompt/icon--arrow-up.svg
adzhindzhi: I think it might be, at least size-wise the stored SVG will be smaller. Though I'm not sure if the difference is significant enough to make a difference.

packages/scratch-gui/src/components/tooltip/tooltip.jsx
kbangelov: Perhaps, but I believe that after bringing out the shared logic for **PopupWithArrow**, the component is simple enough.

packages/scratch-gui/src/components/tooltip/tooltip.css
adzhindzhi: I think we should overall try to avoid negative gaps/margins. It could be something that was added unintentionally (e.g. sometimes Figma adds certain styles when you move components). I think we should either remove it if we don't need it or try to achieve the same UI another way if possible.

packages/scratch-gui/src/components/tooltip/tooltip.jsx
adzhindzhi: +1, most of these useEffects will become redundant if we switch to a Modal.

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
```diff
@@ -0,0 +1,157 @@
...
+    onConfirm,
+    onCancel,
+    isOpen,
+    relativeElementRef,
+    side,
+    align,
+    layoutConfig
+}) => {
+    const {modalWidth, spaceForArrow, counterOffset, arrowOffsetFromBottom, arrowHeight, arrowWidth} =
+        {...defaultConfig, ...layoutConfig};
```
adzhindzhi: nitpick: Let's put each property on a new line for readability.

packages/scratch-gui/src/components/feature-callout-popover/feature-callout-popover.jsx
```diff
@@ -0,0 +1,139 @@
...
+        target.addEventListener('focus', handleFocus);
+        target.addEventListener('blur', handleBlur);
+
+        return () => {
+            target.removeEventListener('mouseenter', handleMouseEnter);
+            target.removeEventListener('mouseleave', handleMouseLeave);
+            target.removeEventListener('focus', handleFocus);
+            target.removeEventListener('blur', handleBlur);
+        };
+    }, [isManualOnly, onRequestOpen, onRequestClose, targetRef?.current]);
```
adzhindzhi: Hmmmm, do we need this logic? For actual tooltips, we use `react-tooltip`, so I'm not sure if there's any case in which we'd want to use this component for that. Also, I find it a bit confusing that we've named this component `Tooltip` (and have also implemented actual tooltip logic), but use a Modal underneath.. I think it'd be cleaner if we removed this logic, since it's currently unused and maybe rename this component to `FeatureCallout` or `FeatureCalloutPopover`. And also remove the `isManualOnly` prop, since this will always be manual.

packages/scratch-gui/src/lib/calculatePopupPosition.js
```diff
@@ -0,0 +1,102 @@
...
+    relativeElementRef,
+    popupRef,
+    side,
+    align = PopupAlign.CENTER,
+    popupWidth,
+    spaceForArrow,
+    arrowHeight,
+    arrowWidth,
+    counterOffset = 5,
+    arrowOffsetFromBottom = 0
```
adzhindzhi: I think adding some docs would be very helpful, especially for properties like `counterOffset`, `spaceForArrow`, etc..

packages/scratch-gui/src/lib/calculatePopupPosition.js
```diff
@@ -0,0 +1,102 @@
...
+}) => {
+    const el = relativeElementRef?.current;
+    const modalEl = popupRef?.current;
+    if (!el || !modalEl) return {};
+
+    const modalHeight = popupRef.current.getBoundingClientRect().height;
+    const buttonRect = el.getBoundingClientRect();
```
adzhindzhi: nitpick: Maybe rename this to targetRect? As it may not always be a button?

packages/scratch-gui/src/components/feature-callout-popover/feature-callout-popover.jsx
```diff
@@ -0,0 +1,139 @@
...
+        width,
+        spaceForArrow,
+        counterOffset,
+        arrowOffsetFromBottom,
+        arrowHeight,
+        arrowWidth
+    } = {...defaultConfig, ...layoutConfig};
+
+    // Simulate hover and focus (normal) tooltip behavior
+    useEffect(() => {
```
KManolov3: Can we remove this useEffect now?

packages/scratch-gui/src/components/feature-callout-popover/feature-callout-popover.jsx
```diff
@@ -0,0 +1,139 @@
...
+    }, [isManualOnly, onRequestOpen, onRequestClose, targetRef?.current]);
+
+    return (
+        <PopupWithArrow
+            isOpen={isOpen}
+            onRequestClose={onRequestClose}
+            relativeElementRef={targetRef}
+            side={side}
+            align={align}
+            layoutConfig={{
```
KManolov3: This object has potential for being memoized - otherwise it's recreated on each render -> the reference changes -> may cause unnecessary rerenders of PopupWithArrow

packages/scratch-gui/src/components/stage-header/stage-header.jsx
kbangelov: I think it's already covered in the **mapStateToProps** in the container component.

packages/scratch-gui/src/components/feature-callout-popover/feature-callout-popover.jsx
```diff
@@ -0,0 +1,139 @@
...
+                popupWidth: width,
+                spaceForArrow,
+                counterOffset,
+                arrowOffsetFromBottom,
+                arrowHeight,
+                arrowWidth
+            }}
+            arrowConfig={arrowConfig}
+        >
+            {({popupRef, pos}) => (
```
KManolov3: It will likely be clearer if this is separated in its own component

packages/scratch-gui/src/components/tooltip/tooltip.jsx
```diff
@@ -0,0 +1,139 @@
...
+};
+
+const arrowConfig = {
+    arrowDownIcon,
+    arrowUpIcon,
+    arrowLeftIcon,
+    arrowRightIcon
+};
+
+const Tooltip = ({
```
KManolov3: Maybe something like `ManualSetThumbnailCalloutModal`?

packages/scratch-gui/src/components/popup-with-arrow/popup-with-arrow.jsx
```diff
@@ -0,0 +1,164 @@
...
+
+    // Update position when isOpen changes
+    useEffect(() => {
+        if (isOpen && popupRef.current && relativeElementRef?.current) {
+            updatePosition();
+        }
+    }, [isOpen, relativeElementRef, updatePosition]);
+
+    // Click outside to close
+    useEffect(() => {
```
KManolov3: Actually, in the case of the feature callout, we wouldn't want to have the "Click outside to close" behaviour. We'll need to make this configurable when we get to that.

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
```diff
@@ -0,0 +1,163 @@
...
+            }}
+            arrowConfig={arrowConfig}
+        >
+            {({popupRef, pos}) => (
+                <ReactModal
+                    isOpen
+                    onRequestClose={onCancel}
+                    contentLabel={title}
+                    style={{
+                        content: {
```
KManolov3: Can we not inline the styles in the jsx?

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
```diff
@@ -0,0 +1,163 @@
...
+                spaceForArrow,
+                counterOffset,
+                arrowOffsetFromBottom,
+                arrowHeight,
+                arrowWidth
+            }}
+            arrowConfig={arrowConfig}
+        >
+            {({popupRef, pos}) => (
+                <ReactModal
```
KManolov3: It seems strange to pass the modal as a child here and then reimplement the modal behaviour in the PopupWithArrow. Maybe we can make it wrap the passed children in a modal by default? That'd hopefully allow us to remove some of the useEffects which reimplement out-of-the-box modal behaviour

packages/scratch-gui/src/components/popup-with-arrow/popup-with-arrow.jsx
```diff
@@ -0,0 +1,164 @@
+import React, {useRef, useState, useCallback, useEffect} from 'react';
+import PropTypes from 'prop-types';
+import calculatePopupPosition, {PopupAlign, PopupSide} from '../../lib/calculatePopupPosition';
+
+const PopupWithArrow = ({
```
KManolov3: Let's make the `DeleteConfirmationPrompt` use that as well.

packages/scratch-gui/src/components/popup-with-arrow/popup-with-arrow.jsx
kbangelov: Should it just remain on screen untill the target ref is clicked?

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
kbangelov: I'm not sure we want every popup with arrow to be a modal?

packages/scratch-gui/src/components/popup-with-arrow/popup-with-arrow.jsx
adzhindzhi: I think if we make the `PopupWithArrow` always a modal (which I don't see a reason not to do), we wouldn't have to handle the click outside in a `useEffect`. Because the modal has properties such as `shouldCloseOnOverlayClick`, `shouldCloseOnEsc`, etc. which we can utilize and block certain actions for the `FeatureCallout` component, which we would eventually want to do, not necessarily as part of this task.

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
KManolov3: Let's assume we do for now, until we reach a case where we don't

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
```diff
@@ -0,0 +1,171 @@
...
+        modalWidth,
+        spaceForArrow,
+        counterOffset,
+        arrowOffsetFromBottom,
+        arrowHeight,
+        arrowWidth
+    } = {...defaultConfig, ...layoutConfig};
+
+    const memoizedLayoutConfig = React.useMemo(() => ({
+        popupWidth: modalWidth,
```
adzhindzhi: nitpick: Can we use either modalWidth or popupWidth everywhere for consistency?

packages/scratch-gui/src/components/delete-confirmation-prompt/delete-confirmation-prompt.jsx
```diff
@@ -87,88 +83,101 @@ const DeleteConfirmationPrompt = ({
...
+    const side = MODAL_POSITION_TO_SIDE[modalPosition] ?? PopupSide.RIGHT;
+
+    return (
+        <PopupWithArrow
+            isOpen
+            onRequestClose={onCancel}
+            relativeElementRef={relativeElementRef}
+            side={side}
+            align={PopupAlign.CENTER}
+            layoutConfig={{
```
KManolov3: Nitpick: we can memoize this

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
adzhindzhi: What I find slightly confusing about the current implementation is the following structure:

- `PopupWithArrow` is a generic component used by `ConfirmationPrompt`, `DeleteConfirmationPrompt`, and `FeatureCallout`.
- `ConfirmationPrompt` is also a generic component, but it's currently only used for the thumbnail confirmation prompt and not for the delete confirmation prompt.

I think what we actually need is a clearer separation of responsibilities:

- `SetTooltipConfirmationPrompt` should use `ConfirmationPrompt` and pass in specific content such as `title` (if needed), `description`, `className`, `layout`, etc.
- `DeleteConfirmationPrompt` should also use `ConfirmationPrompt` and provide its own specific content.
- `ConfirmationPrompt` should use `PopupWithArrow`, which handles alignment and positioning.
- `FeatureCallout` should use `PopupWithArrow` for alignment and positioning.
- `PopupWithArrow` should use `ReactModal` to leverage its built-in behavior for handling outside clicks, escape key presses, etc.

This structure would simplify the overall design and ensure that each component has a clear, single responsibility. We may still need to refine the naming to better reflect each component's purpose, but this is the general direction I think we should take.

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
```diff
@@ -42,18 +43,30 @@ const arrowConfig = {
     arrowRightIcon
 };
 
+const BUTTON_ORDER = {
```
KManolov3, line 46: Ideally, I'd generalize the props to:
````
`mainButtonIcon`
`mainButtonStyles`
`mainButtonLabel`
`mainButtonAction`
`secondaryButtonIcon`
`secondaryButtonStyles`
`secondaryButtonLabel`
`secondaryButtonAction`
````

But I'm not insisting us do that here, as the confirm/cancel separation fits the current cases

File: packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx [hunk 72]
```diff
@@ -42,18 +43,30 @@ const arrowConfig = {
...
+    CANCEL_FIRST: 'cancelFirst',
+    CONFIRM_FIRST: 'confirmFirst'
+};
+
 const ConfirmationPrompt = ({
     title,
     message,
     confirmLabel,
     cancelLabel,
+    confirmIcon,
```
KManolov3: Can we group the related props in a `confirmButtonConfig`, `canceButtonConfig` objects?

packages/scratch-gui/src/components/delete-confirmation-prompt/delete-confirmation-prompt.jsx
```diff
@@ -89,62 +69,25 @@ const DeleteConfirmationPrompt = ({
...
-        arrowWidth
-    }), [modalWidth, arrowWidth, arrowHeight]);
-
     return (
-        <ModalWithArrow
+        <ConfirmationPrompt
             isOpen
-            onRequestClose={onCancel}
+            title={intl.formatMessage(messages.confirmDeletionHeading)}
+            message={<FormattedMessage {...getMessage(entityType)} />}
```
KManolov3, line 76: It looks strange to use FormattedMessage for the message and `intl.formatMessage` for the title. Any reason to have it differ?

packages/scratch-gui/src/components/delete-confirmation-prompt/delete-confirmation-prompt.css
```diff
@@ -53,22 +27,4 @@
     font-size: 0.85rem;
     cursor: pointer;
     margin: auto;
-}
-
-.button-row button.ok-button {
```
KManolov3, line 58: Feelsgood to remove this code

packages/scratch-gui/src/components/delete-confirmation-prompt/delete-confirmation-prompt.jsx
kbangelov: Yes, in **ConfirmationPrompt** I use css styles that I apply to **span**s and **FormattedMessage** renders one. And **title** is just passed purely for accessibility.

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
adzhindzhi: Or alternatively, we can pass `primaryButtonContent`/`secondaryButtonContent` -  a React node, which can either be the text if that's all we want to display, or text + icon. Both seem fine to me, whatever you feel is better.

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
```diff
@@ -0,0 +1,184 @@
...
+        <button
+            onClick={onConfirm}
+            className={classNames(styles.confirmButton, confirmButtonClassName)}
+        >
+            {confirmIcon && (
+                <img
+                    className={styles.buttonIcon}
+                    src={confirmIcon}
+                    aria-hidden="true"
+                    alt=""
```
adzhindzhi: nitpick: Maybe we shouldn't leave the `alt` text empty, since we are also working on making the editor more accessible? Same for the cancel button?

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -69,28 +89,81 @@ const StageHeaderComponent = function (props) {
...
+                            setIsUpdatingThumbnail(false);
+                        }
+                    );
                 });
-            },
-            3000
-        ),
-        [projectId, onUpdateProjectThumbnail]
+            } catch (e) {
+                onShowThumbnailError();
```
adzhindzhi: Would we ever get to the catch of this `try/catch` block? If the thumbnail update fails, it'll call the `onError` callback, but shouldn't fire an error that we can can here? The `storeProjectThumbnail` has its own `try/catch `block, which swallows the error, so it wouldn't bubble up to here?

packages/scratch-gui/src/components/confirmation-prompt/confirmation-prompt.jsx
kbangelov: Since the images are not supposed to be announced, I'll just remove the alt props alltogether.

packages/scratch-gui/src/components/stage-header/stage-header.jsx
adzhindzhi: Can we address this? It generally seems like a bad idea to use `formatMessage` inside a `FormattedMessage`.

packages/scratch-gui/src/reducers/project-state.js
```diff
@@ -100,6 +100,10 @@ const getIsShowingWithId = loadingState => (
 const getIsShowingWithoutId = loadingState => (
     loadingState === LoadingState.SHOWING_WITHOUT_ID
 );
+const getIsProjectLoadedWithId = loadingState => (
+    getIsShowingWithId(loadingState) ||
+    getIsUpdating(loadingState)
+);
```
adzhindzhi: Usually, for a loaded project we're using `getIsShowingWithId`, is there a particular reason why you want to consider an updating project as "loaded"? Wouldn't this also mark new projects that are currently creating as loaded? We're using flag only for whether or not to show the Set Thumbnail button, right? Would it make a difference if we show the button if the project isn't entirely loaded?

packages/scratch-gui/src/components/modal-with-arrow/modal-with-arrow.jsx
```diff
@@ -0,0 +1,164 @@
...
+            window.removeEventListener('resize', updatePosition);
+            window.removeEventListener('scroll', updatePosition, true);
+        };
+    }, [isOpen, updatePosition]);
+
+    const onPopupMount = useCallback(el => {
+        if (!el || !isOpen) return;
+        modalRef.current = el;
+        updatePosition();
+    }, [isOpen, updatePosition]);
```
adzhindzhi, line 86: nitpick: You can probably pass the `modalRef` to the `contentRef` directly and just update the position in the `useEffect` when `isOpen` changes

packages/scratch-gui/src/components/stage-header/stage-header.jsx
```diff
@@ -139,43 +207,79 @@ const StageHeaderComponent = function (props) {
...
                     <div className={styles.stageSizeRow}>
+                        {/* To remove - new feature awareness tooltip */}
+                        <FeatureCalloutPopover
+                            isOpen={isThumbnailTooltipOpen}
+                            onRequestOpen={onOpenTooltip}
+                            onRequestClose={onCloseTooltip}
+                            targetRef={thumbnailButtonRef}
+                            side={PopupSide.LEFT}
+                            align={PopupAlign.DOWN}
+                            layoutConfig={{width: 336}}
```
adzhindzhi: nitpick: This width is the same as the default one in the component, is there a reason why we specify it?

packages/scratch-gui/src/components/feature-callout-popover/feature-callout-popover.jsx
```diff
@@ -0,0 +1,121 @@
...
+    arrowDownIcon,
+    arrowUpIcon,
+    arrowLeftIcon,
+    arrowRightIcon
+};
+
+const FeatureCalloutContent = ({title, body}) => (
+    <Box
+        className={styles.popover}
+        tabIndex={0}
```
adzhindzhi, line 33: nitpick: Do we need this to set the `tabIndex` manually? Does it make sense for the tooltip to be navigable through tab if it's not actionable?

adzhindzhi (approved):
I left a few small comments, but overall it looks good to me! 🙌  I also noticed that the FeatureCalloutPopover is not positioned correctly on initial render, but I think that can be addressed when we start working on the feature callouts.

packages/scratch-gui/src/reducers/project-state.js
kbangelov: As I expected, it causes the icon to reload when we, for example, update the project. I'll replace it with `getIsShowingWithId`, since it seems to happen fast. Although I don't know if there should be a link between saving the project and saving the thumbnail.

packages/scratch-gui/src/components/modal-with-arrow/modal-with-arrow.jsx
kbangelov: I prefer this, because it updates position when the element has already mounted.

packages/scratch-gui/src/components/stage-header/stage-header.jsx
kbangelov: In case we change the default, I guess. Which begs the question whether to pass all the props in the config or not. This one seems most important.

packages/scratch-gui/src/components/feature-callout-popover/feature-callout-popover.jsx
kbangelov: Mmm yes, I think it should be in order to be readable by people who rely entirely on their screen reader. Not sure they'd be able to access it otherwise. Tereza agreed on it.

packages/scratch-gui/src/components/stage-header/stage-header.jsx
adzhindzhi: Well if we at some point need to change the default, we can pass it then, no? It just seems redundant now.

packages/scratch-gui/src/reducers/project-state.js
adzhindzhi: Well, to be fair, in that case I don't really see a valid reason to check whether the project is loading/saved at all. Is there a particular reason why this check was added? If not, maybe we can remove it altogether and just always display the button if the feature flag is on and the user owns the project?

packages/scratch-gui/src/components/modal-with-arrow/modal-with-arrow.jsx
KManolov3: I think we should get rid of the use effect if we can. If the only concern is running `updatePosition` one extra time on mount, then it shouldn't be much of an issue.

KManolov3 (approved):
Looks good to me! 🙌 Some very minor comments left

PR #472, opened 2026-03-13
fix: replace toString with Cast.toString in switch blocks

OceanIsEndless:
Prevents the switch costume to (), switch backdrop to (), and switch backdrop to () and wait blocks from hanging indefinitely in certain situations.

### Resolves

N/A (I would rather not create an issue for this, as it's not a problem that would be faced by users using Scratch as intended.)

### Proposed Changes

This is a super minor change, it just prevents the switch costume to (), switch backdrop to (), and switch backdrop to () and wait blocks from hanging indefinitely in certain situations by changing them to use Cast.toString() instead of .toString when converting the input to a string.

### Reason for Changes

To be more consistent with other blocks.

As far as I can tell, all other blocks in the runtime do not rely on the .toString method being available on values passed as input. These blocks do, so if they're given a value without a .toString method, they will throw an error (after which the blocks are run again, and again, and again, throwing the same error and never continuing the script). Changing it to use Cast.toString() makes it more consistent with other blocks. While for most intents and purposes this will never be an issue, it will prevent it from failing under extraordinary circumstances, e.g. if it is given `undefined` as input, which does not have a .toString method.

### Test Coverage

N/A? I'll add a test for it if it's absolutely necessary, but the behavior is effectively the same as before, it just does not throw when given an invalid value.

### DIFF
--- a/packages/scratch-vm/src/blocks/scratch3_looks.js
+++ b/packages/scratch-vm/src/blocks/scratch3_looks.js
@@ -392,7 +392,7 @@ class Scratch3LooksBlocks {
             target.setCostume(optZeroIndex ? requestedCostume : requestedCostume - 1);
         } else {
             // Strings should be treated as costume names, where possible
-            const costumeIndex = target.getCostumeIndexByName(String(requestedCostume));
+            const costumeIndex = target.getCostumeIndexByName(Cast.toString(requestedCostume));
 
             if (costumeIndex !== -1) {
                 target.setCostume(costumeIndex);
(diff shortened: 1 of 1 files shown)


OceanIsEndless:

OceanIsEndless:
Apologies for the force pushes, they were to make the commit message meet commitlint requirements, but now of course the scratch-gui check has failed for no good reason. 😕

Edit: The issue was fixed. :)

yuri-kiss:
You should probably use `Cast.toString` because that's the formality

OceanIsEndless:
> You should probably use `Cast.toString` because that's the formality

Oh thanks, somehow missed that, though I wonder why it didn't do this in the first place...

yuri-kiss:
@cwillisf could you look at this :pleading_face:

yuri-kiss:

PR #478, opened 2026-03-16
Publish tested builds

cwillisf:
### Proposed Changes

Main change: make sure that the publish workflow uses the same build output that was tested.

Supporting changes: lots of GHA cleanup & refinement.

### Reason for Changes

This removes the possibility of introducing new issues between the build-and-test in the CI workflow and the build-and-publish of the publishing workflow. In theory, those two builds _should_ be the same, but we couldn't be certain of that before. By publishing the existing build artifacts, we can now be certain. It also saves some time in the publishing workflow. In the event that we publish a commit-ish for which no build is available - for example, if we publish a commit that affected only docs, or only files in `.github/` - then fall back to building everything as part of the publishing workflow.

The rest of the changes are more of a "while I'm in here" set of things, including:

- Standardize on running values through `env:` for [these reasons](https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable). Not all of the values are vulnerable to that, but I'd like to establish it as a habit.
- Skip both `test` and `results` if no packages have been modified
- Skip at the GHA level instead of the bash level when possible
- When a new release is published, add comments on issues and PRs associated with the release

### Test Coverage

I tested as much as I could locally by running shell commands in a terminal, etc., but the real test will be when this is merged and the first time we publish a new release after that.


.github/workflows/publish.yml
```diff
@@ -5,42 +5,105 @@ on:
...
+        shell: bash
+        run: |
+          # Resolve branch name or tag to a commit SHA
+          SHA=$(gh api "repos/$GITHUB_REPOSITORY/commits/$TARGET" --jq '.sha' 2>/dev/null || echo "$TARGET")
+          echo "Resolved target '$TARGET' to SHA: $SHA"
+
+          # Find the most recent successful CI run for this SHA
+          RUN_ID=$(gh api "repos/$GITHUB_REPOSITORY/actions/workflows/ci.yml/runs" \
+            -F "head_sha=$SHA" \
+            -F "status=success" \
```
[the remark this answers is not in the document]
cwillisf: Checking [this API reference](https://raw.githubusercontent.com/github/rest-api-description/main/descriptions/api.github.com/api.github.com.json), specifically the `/repos/{owner}/{repo}/actions/workflows/{workflow_id}/runs` endpoint, I think Copilot got confused. The `status=success` here is a `workflow-run-status` value, whereas Copilot seems to be thinking of the `status` field of some other request (for example, maybe `/repos/{owner}/{repo}/check-runs`).

.github/workflows/publish.yml
```diff
@@ -141,9 +208,19 @@ jobs:
             echo "Not pushing to develop because the tag we're operating on is behind"
           fi
 
+      - name: Comment on resolved issues and merged PRs
```
KManolov3, line 218: Cool!

.github/workflows/publish.yml
```diff
@@ -63,76 +139,67 @@ jobs:
...
 
       - name: Publish scratch-gui
         run: |
           cp ./packages/scratch-gui/package.json ./packages/scratch-gui/package-copy.json
 
           jq 'del(.exports["./standalone"])' ./packages/scratch-gui/package.json | npx sponge ./packages/scratch-gui/package.json
 
-          npm run build:dist --workspace @scratch/scratch-gui
-          npm publish --access=public --tag="${{steps.npm_tag.outputs.npm_tag}}" --workspace=@scratch/scratch-gui
+          npm publish --access=public --tag="$NPM_TAG" --ignore-scripts --workspace=@scratch/scratch-gui
```
KManolov3, line 183: I wondered why we were building here as well. I remember an issue from a while back, where GUI was using the changes from the old published version of svg-renderer/vm/.. instead of what's locally available in the build. 
The second `install-dependencies` that you removed referred to that. I suppose that since the build is now done before updating the version, it shouldn't have the same issue - is that assumption correct?

KManolov3 (approved):
Looks good to me! Pretty happy that we'll be cutting the publish time 🙌

.github/workflows/publish.yml
cwillisf: Honestly, I'm a bit nervous about things like that. I *think* it'll be OK, since all the `package.json` files are updated by the `npm version` command calling `scripts/npm-version.sh`, and I don't believe the `package.json` files are built into the bundle output. If I'm wrong about that, then this needs to be more complicated. (It probably should be complicated, but I was hoping to take it one step at a time...)

I think the ideal is something like:
1. make a local version bump commit
2. build and test
3. if tests pass, merge the version bump upstream (FF if possible)
4. npm publish

I'm still working through that mentally... let me know if you see any gaps. The build scripts for our Android app do something like that, and it's a bit of a pain to set up, but it works.

Coming back to reality, I will definitely be carefully inspecting the first publish after this merge. It wouldn't surprise me too much if there's a problem introduced by the build happening before the version bump, in which case I'll either try to fix it quickly or revert this change.

PR #489, opened 2026-03-22
fix: pick random scientific notation

### LINKED ISSUE #160: `pick random () to ()` integer(‐ish) output with scientific notation
The `pick random () to ()` block detects whether it should output only integers by checking whether either of its inputs contains a dot, but it’s possible to input fractional numbers in a different way, i.e. with scientific notation.

E.g. if you enter `5e-1` in its first field (and leave the second field as `10`, it will “attempt” to output only integers, but they will be offset by 0.5 (outputting e.g. 1.5, 6.5, 3.5, etc.)

The expected/intuitive behavior is that it should output an arbitrary float value in the range just as if `0.5` had been entered.

`pick random (5e-1) to (10)` outputting 3.5
`pick random (5e-1) to (10)` outputting 3.285...

wusctong:
### Resolves
Fix #160

### Proposed Changes
- Updated `Cast.isInt` string branch to handle scientific notation: if the string contains an `e`, parse it numerically to check if it's truly a whole number, otherwise fall back to the dot check.
- Updated the existing test for `'0.1e10'` from `false` to `true`, since `0.1e10` equals `1000000000` which is a whole number.

### Reason for Changes
`Cast.isInt` only checked for a `.` to decide if a string was an integer. This broke with scientific notation — `5e-1` has no dot but is `0.5` (a float), causing `pick random (5e-1) to (10)` to return offset values like `1.5`, `3.5` instead of proper floats.

### Test Coverage
Updated `util_cast.js` — changed `cast.isInt('0.1e10')` expected value from `false` to `true`, since `0.1e10` (`1000000000`) is a whole number and should be treated as an int.

### DIFF
--- a/packages/scratch-vm/src/util/cast.js
+++ b/packages/scratch-vm/src/util/cast.js
@@ -167,7 +167,7 @@ class Cast {
             return true;
         } else if (typeof val === 'string') {
             // If it contains an 'e' (scientific notation), parse it numerically to check if it's an int.
-            if (val.toLowerCase().split('e').length <= 2) {
+            if (val.toLowerCase().split('e').length === 2) {
                 return Number(val) === parseInt(Number(val), 10);
             } else {
                 // If it contains a decimal point, don't consider it an int.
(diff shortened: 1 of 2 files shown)


wusctong:

PR #492, opened 2026-03-23
[UEPR-527] Investigate membership sprites rendering differently in paint editor vs. stage

kbangelov:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-527

### Proposed Changes

Explicitly handles logic for "**clip-path**" elements which otherwise don't pass down **transform** to its children, thus their disappearance. Big credit to Claude for this one.

### DIFF
--- a/packages/scratch-svg-renderer/src/transform-applier.js
+++ b/packages/scratch-svg-renderer/src/transform-applier.js
@@ -516,7 +516,7 @@ const transformStrokeWidths = function (svgTag, windowRef, bboxForTesting) {
     const _applyTransformToClipPath = function (element, matrix) {
         const clipPathAttr = element.attributes && element.attributes['clip-path'];
         if (!clipPathAttr) return;
-        const clipPathId = _parseUrl(clipPathAttr.value, null);
+        const clipPathId = _parseUrl(clipPathAttr.value, windowRef);
         if (!clipPathId) return;
         const clipPathEl = svgTag.getElementById(clipPathId);
         if (!clipPathEl) return;
--- a/packages/scratch-svg-renderer/src/transform-applier.js
+++ b/packages/scratch-svg-renderer/src/transform-applier.js
@@ -487,6 +487,20 @@ const _parseUrl = (value, windowRef) => {
     return res;
 };
 
+// Fixes an issue where clip paths didn’t follow transforms.
+//
+// In SVG, elements can have transforms (scale, rotate, translate, etc.)
+// applied via a transform matrix. In this file, we apply those transforms
+// into the actual geometry (e.g. path data) instead of keeping them as
+// separate transform attributes.
+//
+// However, clip paths are defined separately and are referenced by elements.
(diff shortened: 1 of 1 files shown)


packages/scratch-svg-renderer/src/transform-applier.js
```diff
@@ -512,6 +512,17 @@ const _parseUrl = (value, windowRef) => {
 const transformStrokeWidths = function (svgTag, windowRef, bboxForTesting) {
     const inherited = Matrix.identity();
 
+    // pass down transforms to leaf level
+    const _applyTransformToClipPath = function (element, matrix) {
```
KManolov3, line 565: Why did the SVG (e.g. Mia's eyes) render correctly in the canvas, but not in the stage?

kbangelov: It probably has something to do with the transformation recalculating the gradient relative to a new bounding box after transforms are applied. The reason it goes wrong is because the code for the eye on the right

`<linearGradient id="av" x1="-1309.38" y1="83.32" x2="-1294.744" y2="73.781"
  gradientTransform="translate(-1196.214) rotate(-180) scale(1 -1)"
  xlink:href="#as"/>`

is missing 

`gradientUnits="userSpaceOnUse"`, which we do need to set here as well in order to correctly apply the colors at the right coordinates.

packages/scratch-svg-renderer/src/transform-applier.js
```diff
@@ -523,12 +534,15 @@ const transformStrokeWidths = function (svgTag, windowRef, bboxForTesting) {
                 if (element.attributes.stroke) stroke = element.attributes.stroke.value;
             }
 
+            const currentMatrix = Matrix.compose(matrix, _parseTransform(element));
+            _applyTransformToClipPath(element, currentMatrix);
+
```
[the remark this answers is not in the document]
kbangelov: I think we already have similar tests.

packages/scratch-svg-renderer/src/transform-applier.js
adzhindzhi: Hm, so in general what we want to do is calculate the transform matrix from the `<g>` element, where the `clip-path` attribute is defined, then find that `clipPath` and pass the transform down to it? Do you have any idea why this resolves the issue? I just wonder if Copilot has a point about this having to be handled in a similar way to the gradients, where as far as I can see the matrix is passed as part of the url, rather than being attached to the `linearGradient` element directly and also the path `d` attribute is transformed (as for the stroke width).

kbangelov: I believe the **clipPath** is not considered a child and never gets the transform passed down to it. But yes, perhaps it does have a point. I just pushed a commit to add logic to stop modifying original clip path and create new instead, similarly to gradient.

KManolov3 (approved):
Looks good!

adzhindzhi (approved):

PR #496, opened 2026-03-24
fix(ci): build with new version before testing and publishing

cwillisf:
### Resolves

Resolves an issue I introduced with scratchfoundation/scratch-editor#478

### Proposed Changes

The main change is to adjust `publish.yml` to perform these steps in this order (assuming no failures):

1. Check out
2. Apply new version number
3. Build
4. Test
5. Publish

As opposed to the behavior last week:

1. Check out
2. Build
3. Test
4. Apply new version number
5. Build
6. Publish

Or the behavior introduced by scratchfoundation/scratch-editor#478:

1. Check out
2. Build
3. Test
4. Apply new version number
5. Publish the build from step 2, but with the adjusted version metadata

Supporting changes: significant refactoring of `ci.yml` and `publish.yml`, including the creation of separate (called) workflows for testing and publishing to GH Pages.

### Reason for Changes

After merging scratchfoundation/scratch-editor#478, I discovered that packages are published under their intended version number, but _contain_ the previous version number. For example, if you inspect `dist/scratch-gui.js` from [@scratch/scratch-gui@13.4.0-test.3](https://www.npmjs.com/package/@scratch/scratch-gui/v/13.4.0-test.3), you'll notice the version number `13.3.0` embedded within. The changes here fix that. (It's too bad nobody could have [predicted](https://github.com/scratchfoundation/scratch-editor/pull/478#discussion_r2976096516) this problem! :sweat_smile:)

The ~easiest~ natural way to fix that meant rearranging the existing steps so much that calling `ci.yml` from `publish.yml` no longer made sense, so I refactored the workflows to keep things somewhat DRY.

As a bonus, this means `publish.yml` can build everything without worrying about path filters, and lays the groundwork for matrixing the build step in `ci.yml`.

### Test Coverage

Partially tested locally, but I'll also do a test publish or twelve after merging to resolve any issues that are difficult to predict.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,81 +5,11 @@ on:
     types: [published]
 
 jobs:
-  find-artifacts:
-    name: Find CI artifacts
-    runs-on: ubuntu-latest
-    permissions:
-      actions: read
-      contents: read
-    outputs:
-      ci-run-id: ${{ steps.find.outputs.run-id }}
-      needs-build: ${{ steps.find.outputs.needs-build }}
-    steps:
-      - name: Find successful CI run for release target
-        id: find
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          TARGET: ${{ github.event.release.target_commitish }}
-        shell: bash
-        run: |
-          # Resolve branch name or tag to a commit SHA
-          SHA=$(gh api "repos/$GITHUB_REPOSITORY/commits/$TARGET" --jq '.sha' 2>/dev/null || echo "$TARGET")
-          echo "Resolved target '$TARGET' to SHA: $SHA"
-
-          # Find the most recent successful CI run for this SHA that has a build artifact
-          RUN_ID=$(gh api "repos/$GITHUB_REPOSITORY/actions/workflows/ci.yml/runs" \
-            --method GET \
-            -F "head_sha=$SHA" \
-            -F "status=success" \
-            -F "per_page=10" \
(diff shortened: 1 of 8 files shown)


.github/workflows/publish.yml
```diff
@@ -5,95 +5,88 @@ on:
...
+          NEW_VERSION="${GIT_TAG/v/}"
+          npm version "$NEW_VERSION" --no-git-tag-version
 
-  build:
-    name: Build (no prior CI run found)
-    needs: find-artifacts
-    if: ${{ needs.find-artifacts.outputs.needs-build == 'true' }}
-    uses: ./.github/workflows/ci.yml
+      - name: Build Packages
+        run: npm run build
```
KManolov3, line 44: Does `npm run build` guarantee us the packages are built in the correct order (render/vm.. before GUI which uses them)? I assume yes, since we haven't seen any such errors in the GH runs.. though I do wonder - I have a hard time finding anything concrete on whether `cross-env NODE_ENV=production npm run --workspaces build` guarantees the build order according to the dependency graph.

KManolov3 (approved):
Looks good to me! 🤞

.github/workflows/publish.yml
cwillisf: When you use `--workspaces` to tell `npm` to go through all the workspaces, it goes through them in the order specified in the `package.json` (or it picks the order if you use wildcards). The `scratch-editor` workspaces are currently ordered manually to make sure the dependency graph is respected.

PR #505, opened 2026-03-27
Feat/uepr 535 introduce script for migrating packages to monorepo

KManolov3:
TODO:
- **Test more extensively for corner cases**

### Resolves

Part of https://scratchfoundation.atlassian.net/browse/UEPR-535

### Proposed Changes

Add two scripts - `add-repo.sh` and `refresh-gh-workflows` to assist in adding new packages to monorepo.

### Reason for Changes

We want to expand the scratch-editor monorepo with other packages that make sense to be there (e.g. scratch-paint, scratch-audio, ...)

### Test Coverage

This is for development use, and the results of it should always undergo manual developer review, so I'm unsure if automated tests make sense in terms of effort.


.github/workflows/publish.yml
```diff
@@ -192,11 +192,7 @@ jobs:
         run: npm publish --access=public --tag="$NPM_TAG" --ignore-scripts --workspace=@scratch/task-herder
 
       - name: Publish scratch-media-lib-scripts
-        run: |
```
KManolov3, line 195: Those were some leftovers from the publish update PR, the build and token set happen earlier currently.

scripts/add-repo.sh
```diff
@@ -0,0 +1,467 @@
...
+    done
+done
+
+# Replace require/import references to the new repo's un-prefixed name across the codebase
+echo "==> Updating require/import references for ${REPO_NAME}..."
+find "${MONOREPO_ROOT}" -type f \
+    -not -path '*/.git/*' \
+    -not -path '*/node_modules/*' \
+    -exec grep -Il "$REPO_NAME" {} \; 2>/dev/null | while read -r file; do
+    sed -i '' -e "s:\(require(\|from \|resolve(\|node_modules\)\(['\"/]\)${REPO_NAME}\(['\"/]\):\1\2${NPM_ORGANIZATION}/${REPO_NAME}\3:g" "$file"
```
kbangelov, line 400: `sed -i '' -e` seems to be a macOS-only syntax. Consider using `perl -pi -e`

scripts/add-repo.sh
```diff
@@ -0,0 +1,467 @@
...
+    fi
+
+    # Disconnect from the reference repo
+    git -C "${BUILD_TMP}/${repo}" repack -a
+    rm -f "${BUILD_TMP}/${repo}/.git/objects/info/alternates"
+}
+
+# Rewrite all paths in the cloned repo to live under a subdirectory
+# Handles repos with and without submodules
+move_repository_subdirectory() {
```
KManolov3, line 179: This was taken from the build-monorepo script. We don't (yet) have a case of repo with submodules.

scripts/add-repo.sh
```diff
@@ -0,0 +1,467 @@
...
+            'del(.devDependencies."@commitlint/cli")' \
+            'del(.devDependencies."@commitlint/config-conventional")' \
+            'del(.devDependencies."@commitlint/travis-cli")' \
+            'del(.devDependencies."cz-conventional-changelog")' \
+            'del(.devDependencies."husky")' \
+            'if .devDependencies == {} then del(.devDependencies) else . end' \
+        ) "${PACKAGE_PATH}/package.json" | sponge "${PACKAGE_PATH}/package.json"
+fi
+
+### Step 5: Update root workspaces list ###
```
KManolov3, line 312: Consider using `npm init` as an alternative (as in https://github.com/scratchfoundation/scratch-editor/pull/434/changes#diff-c0835ea4a8034dbcc4211a854aa9ee66ca48777eca6294856e5e02e60ad0b9fdR1-R358)

scripts/add-repo.sh
```diff
@@ -0,0 +1,467 @@
...
+            SOURCE_BRANCH="$2"
+            shift 2
+            ;;
+        --org)
+            GITHUB_ORG="$2"
+            shift 2
+            ;;
+        --cache-dir)
+            BUILD_CACHE="$2"
+            shift 2
```
kbangelov, line 63: Perhaps we should consider adding checks for whether the second argument exists and whether it is not another flag, which would lead to nonsense instead of error.

```
if [ -z "$2" ] || [[ "$2" == --* ]]; then
      echo "Error: --source-branch requires a value" >&2
      usage 1
  fi
```

scripts/add-repo.sh
```diff
@@ -0,0 +1,467 @@
...
+
+if ! command -v jq &> /dev/null; then
+    echo "Error: Please install jq. You can try with:" >&2
+    echo "- brew install jq" >&2
+    echo "- sudo apt install jq" >&2
+    exit 1
+fi
+
+if [ -d "$PACKAGE_PATH" ]; then
+    echo "Error: ${PACKAGE_DIR} already exists in the monorepo." >&2
```
kbangelov, line 123: Nice coverage!

scripts/add-repo.sh
```diff
@@ -0,0 +1,467 @@
...
+    for DEP in $OPTDEPS; do
+        npm install --force --save-optional --save-exact \
+            "${NPM_ORGANIZATION}/${DEP}" -w "${NPM_ORGANIZATION}/${PACKAGE}" \
+            || package_replacement_error "$PACKAGE" "$DEP"
+    done
+    for DEP in $PEERDEPS; do
+        npm install --force --save-peer --save-exact \
+            "${NPM_ORGANIZATION}/${DEP}" -w "${NPM_ORGANIZATION}/${PACKAGE}" \
+            || package_replacement_error "$PACKAGE" "$DEP"
+    done
```
kbangelov, line 391: I wonder if this might lead to a situation where the `npm install` fails but `package_replacement_error` just logs the error and continues.

kbangelov (commented):
Looks good to me overall! Left a few small comments.

.github/workflows/publish.yml
adzhindzhi: Thank you!

scripts/add-repo.sh
```diff
@@ -0,0 +1,467 @@
...
+
+# Report that replacing a dependency with the local monorepo version failed
+# $1: the name of the repository
+# $2: the branch that was being built
+# $3: the dependency that failed to install
+package_replacement_error () {
+    echo "***ERROR***"
+    echo "Could not replace a dependency with the local monorepo version."
+    echo "Failed to replace $3 in $1#$2" | tee -a "monorepo.errors.log"
+    #exit 1 # uncomment this to make it a fatal error
```
adzhindzhi, line 229: nicpick: Should we actually uncomment this? Or maybe make it a flag that is passed to the script? Because I think right now we could get to a point where something fails to install correctly, but the script still succeeds. Which makes it easier for the person running the script to miss it?

adzhindzhi (approved):
Looks good to me!

PR #509, opened 2026-03-31
feat: add feature callout for set thumbnail button in editor

KManolov3:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-522

### Proposed Changes

Adjust logic for visualisation of feature callout for the manual thumbnail button, which now lives in the editor

### Reason for Changes

Part of the effort to move the manual thumbnail update inside the editor (as opposed to the project page)

### DIFF
--- a/packages/scratch-gui/src/components/feature-callout-popover/feature-callout-popover.css
+++ b/packages/scratch-gui/src/components/feature-callout-popover/feature-callout-popover.css
@@ -33,3 +33,28 @@
     font-weight: 400;
     line-height: 1.25rem;
 }
+
+.non-blocking-overlay {
+    position: fixed;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
+    z-index: 510;
+    background-color: transparent;
+    pointer-events: none;
+}
+
+.non-blocking-content {
+    border: none;
+    border-radius: 0;
+    height: fit-content;
+    background-color: transparent;
+    padding: 0;
+    margin: 0;
+    position: fixed;
+    overflow: visible;
+    z-index: 1000;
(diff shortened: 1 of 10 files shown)


packages/scratch-gui/src/components/feature-callout-popover/feature-callout-popover.jsx
adzhindzhi, line 15: By the way, the whole popover is wide `21rem`/`336px` (defined in the `.popover` class). However, it's not a `border-box`, so the padding is not included in the width and the modal is actually wider, causing it to get misplaced (left property is not calculated correctly).

adzhindzhi (approved):
Looks good! 🙌

packages/scratch-gui/src/components/feature-callout-popover/feature-callout-popover.css
```diff
@@ -33,3 +33,28 @@
...
+    height: fit-content;
+    background-color: transparent;
+    padding: 0;
+    margin: 0;
+    position: fixed;
+    overflow: visible;
+    z-index: 1000;
+    pointer-events: auto;
+    outline: none;
+}
```
kbangelov: Since there seems to be some code repetition between this and **modal-with-arrow.css**, do you think it might be worth refactoring the component so that it accepts the className on top of the default styles? I'm leaning towards a "no" for the benefit of code clarity here. Just wondering what you think.

kbangelov (commented):
Looks good to me! Really like how this turned out! I'll approve once we clear out the visual positioning mismatch.

kbangelov (approved):

kbangelov (approved):

PR #514, opened 2026-04-07
[UEPR-558] Set thumbnail callout should not be displayed above profile menu

kbangelov:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-558

### Proposed Changes

Adds new z-index with value 490 (1 less than menu-bar's z-index)

### DIFF
--- a/packages/scratch-gui/src/components/feature-callout-popover/feature-callout-popover.css
+++ b/packages/scratch-gui/src/components/feature-callout-popover/feature-callout-popover.css
@@ -66,6 +66,6 @@
 }
 
 .arrow {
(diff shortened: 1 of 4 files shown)


packages/scratch-gui/src/components/feature-callout-popover/feature-callout-popover.css
```diff
@@ -64,3 +65,7 @@
     outline: none; 
 }
 
+.arrow {
+    z-index: $z-index-tooltip !important; /* otherwise it gets overridden in modal-with-arrow */
+}
```
adzhindzhi, line 70: Maybe we can make the selector more specific to avoid using `!important`?

KManolov3 (approved):

PR #515, opened 2026-04-07
[UEPR-552] Add READ_ONLY_MODE flag

kbangelov:
### Resolves

Part of https://scratchfoundation.atlassian.net/browse/UEPR-552

### Proposed Changes

Receives the prop from www


kbangelov:
Closed because we can just use function-specific flags instead. Which I will do in another task.

PR #524, opened 2026-04-14
chore(deps): audit deps, move several to devDependencies

cwillisf:
### Proposed Changes

Three categories of changes:
- A few deps were unused and have been removed
- Some webpack loaders are used only during build and have moved to `devDependencies`
- `scratch-render` exposes two loaders to library consumers when those consumers use the `webpack` export specifically. Those loaders were listed as `dependencies`, but it's more correct to list them as `devDependencies` (for `scratch-render` build) as well as optional `peerDependencies` (for webpack consumers)

### Test Coverage

Covered by the existing build & test steps


cwillisf:
I want to think more carefully about the `peerDependencies` situation before merging this. Also, this should probably be marked as a breaking change due to the `scratch-render` loaders at least, and maybe the `react-redux` change (depending on how I/we decide to handle the `peerDependencies` question).

PR #536, opened 2026-04-20
fix(scratch-vm): repair broken shadow references from cursed inputs bug

cwillisf:
### Resolves

- Crash when dragging reporters out of inputs in corrupted projects (TypeError: Cannot set properties of undefined (setting 'parent') in `moveBlock`)
- Blank/dead inputs in projects affected by the cursed inputs bug (forum topics 878291, 878962)
- Companion to scratchfoundation/scratch-blocks#3560

### Proposed Changes

Projects corrupted by the shared-shadow-ID bug (fixed in scratch-blocks#3560) contain inputs with shadow references that either point to nonexistent blocks or are null when they shouldn't be. This causes crashes when dragging reporters and perpetuates the corruption through save/load cycles. These changes repair corrupted projects at load time and prevent crashes on any remaining edge cases.

**Load-time repair in `deserializeBlocks`** (`sb3.js`): After deserialization, scans all inputs for two forms of shadow corruption:
- **Broken reference:** `shadow` points to a block ID not present in the project
- **Missing shadow:** `shadow` is null when peers of the same opcode have a shadow on the same input

In both cases, recreates the shadow using peer lookup (finds another block of the same opcode with an intact primitive shadow on the same input). Falls back to a `text` shadow if no peer is available (broken references only). Leaves the input alone if no peer has a shadow either (the input probably doesn't use one, e.g. statement inputs like SUBSTACK).

All primitive shadow types are supported, including variable/list/broadcast menu shadows (`data_variable`, `data_listcontents`, `event_broadcast_menu`) — these are created with the correct `id` and `variableType` fields copied from the peer template, or with empty defaults if no peer is available.

**Defensive guard in `moveBlock`** (`blocks.js`): When restoring a shadow after a reporter is disconnected, checks that the shadow block actually exists before accessing it. If missing, clears the stale reference instead of crashing.

**Performance** (`sb3.js`): Peer shadow lookups are cached per `(opcode, inputName)` pair so the blocks map is only scanned once per combination, avoiding O(n^2) cost on projects with many statement inputs.

### Test Coverage

- `engine_blocks.js`: Test that `moveBlock` doesn't crash when shadow reference points to nonexistent block, and that the stale reference is cleared
- `serialization_sb3.js`: Tests for broken shadow reference repair (with peer lookup and `text` fallback), intact shadow preservation, math input peer lookup (`math_number` from peer, not `text`), and null-shadow repair when peers indicate a shadow should exist

### Manual Testing

- [ ] Load project 1296008268. Switch to "Player Costumes" sprite. Find the "Player X Momentum < 0" comparison and pull the reporter out of the block's left operand. Verify the text input appears (not blank).
- [ ] In the same project, pull reporters out of various operator and "set variable to" blocks. Verify inputs are restored in all cases.
- [ ] Verify no crash (was: TypeError in `moveBlock`) when dragging reporters around in corrupted projects.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1919,21 +1919,6 @@
         "specificity": "bin/cli.js"
       }
     },
-    "node_modules/@bramus/specificity/node_modules/css-tree": {
-      "version": "3.1.0",
-      "license": "MIT",
-      "dependencies": {
-        "mdn-data": "2.12.2",
-        "source-map-js": "^1.0.1"
-      },
-      "engines": {
-        "node": "^10 || ^12.20.0 || ^14.13.0 || >=15.0.0"
-      }
-    },
-    "node_modules/@bramus/specificity/node_modules/mdn-data": {
-      "version": "2.12.2",
-      "license": "CC0-1.0"
-    },
     "node_modules/@colors/colors": {
       "version": "1.5.0",
       "dev": true,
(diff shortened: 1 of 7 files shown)


redspacecat:
when will this fix be deployed to the website? 👀

cwillisf:
I meant to get it out today, but I had to leave before I could do that (I volunteer at a place on Monday afternoons). I have lots of meetings on Tuesdays, but I should be able to sneak out a hotfix at some point - by 3 pm US Pacific / 6 pm US Eastern, at least. That's the _current_ plan, of course. Things can always change, especially if someone finds a problem with this fix :sweat_smile:

PR #542, opened 2026-04-24
fix(scratch-vm): clone shared shadows when restoring inputs

Joeclinton1:
### Resolves

Part of #533. Fixes the P0 item where duplicated blocks can still share obscured shadow IDs in some paths, causing duplicated blocks to become corrupted after a reporter is removed from one copy.

Follow-up to #536.

### Proposed Changes

When `moveBlock` restores an obscured shadow after a reporter is disconnected, check whether the referenced shadow block is already parented to a different block. If it is, clone the shadow with a fresh ID for the input being restored instead of reparenting the shared shadow away from the other block.

This keeps each parent input owning its own shadow block and prevents the shared-shadow state from producing top-level/unrenderable shadow blocks after target switches or workspace reloads.

Also logs the repaired state with the relevant block, parent, input, shadow, and shadow parent IDs.

### Reason for Changes

Duplicated blocks affected by the shared-shadow-ID bug can have two inputs pointing at the same obscured shadow block. Removing the inserted reporter from one duplicate used to restore that shared shadow by changing its `parent`, which stole it from the other duplicate and left the VM block graph inconsistent.

Scratch cannot render top-level or incorrectly owned shadow blocks reliably, and this can prevent later scripts from rendering from the project JSON until save/reload removes the bad shadow.

### Test Coverage

- Added `engine_blocks.js` regression coverage for a corrupted shared-shadow state where two parent inputs reference the same obscured shadow ID.
- Verified that removing one reporter clones the shared shadow for that parent, leaves the original parent/shadow relationship intact, and disconnects the reporter cleanly.
- Ran `npx.cmd tap test/unit/engine_blocks.js` from `packages/scratch-vm`.
- Ran `git diff --check`.
- Ran `npx.cmd eslint src/engine/blocks.js test/unit/engine_blocks.js` from `packages/scratch-vm` (0 errors; existing JSDoc warnings in `blocks.js`).

### Manual Testing

- Reproduced the #533 flow locally before the fix: put a block inside another block's input, duplicate both, remove the inner block from one copy, switch sprites, then return.
- Verified locally after the fix that the duplicated block no longer becomes corrupted by the shared shadow ID path.

### DIFF
--- a/packages/scratch-vm/src/engine/blocks.js
+++ b/packages/scratch-vm/src/engine/blocks.js
@@ -9,6 +9,7 @@ const BlocksRuntimeCache = require('./blocks-runtime-cache');
 const log = require('../util/log');
 const Variable = require('./variable');
 const getMonitorIdForBlockWithArgs = require('../util/get-monitor-id');
+const uid = require('../util/uid');
 
 /**
  * @file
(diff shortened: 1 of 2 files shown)


Joeclinton1:
This PR was made in under 14 minutes as speedrun test of GPT 5.5 on codex.

Joeclinton1:
I see this is a duplicate ah

PR #553, opened 2026-04-29
[UEPR-553] Front-end blocking in editor

kbangelov:
### Resolves

Part of https://scratchfoundation.atlassian.net/browse/UEPR-553

### Proposed Changes

Disables project sharing and limits actions for shared projects.


packages/scratch-gui/src/components/gui/gui.jsx
```diff
@@ -139,6 +139,7 @@ const GUIComponent = props => {
         canSave,
         canCreateCopy,
         canShare,
+        canUpdateThumbnail,
```
adzhindzhi, line 142: Let's not do that, we already use and pass the `shouldShowManualThumbnailUpdate` prop based on the `READ_ONLY_MODE` env variable in www and the state of the project. And we agreed that it's okay to not show the thumbnail update button at all in readonly mode, rather than disabling it.

Let's close this PR, and remove the `canUpdateThumbnail` prop from the PR in www.

PR #563, opened 2026-05-08
[UEPR-577] Fix Save Now button not focusable

adzhindzhi:
### Resolves

[UEPR-577](https://scratchfoundation.atlassian.net/browse/UEPR-577)

### Proposed Changes

- Make Save Now button an actual button
- Apply consistent styling with the other menu buttons (changes background on hover)
- Fix issue with `Project Saved.` alert having `cursor: pointer` (due to it being inherited from `menu-item`, where it was recently added)

### Reason for Changes

- Fix issue with Save Now button not focusable on accessibility branch

[UEPR-577]: https://scratchfoundation.atlassian.net/browse/UEPR-577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/scratch-gui/src/components/menu-bar/save-status.css
+++ b/packages/scratch-gui/src/components/menu-bar/save-status.css
@@ -1,6 +1,5 @@
 .save-now {
     font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
-    cursor: pointer;
 }
 
 .alert {
--- a/packages/scratch-gui/src/components/menu-bar/save-status.css
+++ b/packages/scratch-gui/src/components/menu-bar/save-status.css
@@ -2,3 +2,7 @@
     font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
     cursor: pointer;
 }
+
+.alert {
+    cursor: default;
+}
--- a/packages/scratch-gui/src/components/alerts/inline-message.jsx
+++ b/packages/scratch-gui/src/components/alerts/inline-message.jsx
@@ -10,10 +10,11 @@ import styles from './inline-message.css';
 const InlineMessageComponent = ({
     content,
     iconSpinner,
-    level
+    level,
+    className
 }) => (
     <div
-        className={classNames(styles.inlineMessage, styles[level])}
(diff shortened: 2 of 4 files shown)


kbangelov (approved):
Approve.

packages/scratch-gui/src/components/menu-bar/save-status.css
```diff
@@ -2,3 +2,7 @@
     font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
     cursor: pointer;
 }
+
+.alert {
+    cursor: default;
```
kbangelov, line 6: Minor nitpick: Not sure what this does. 

Additional nitpick: I can notice that adding the menuBarItem and hoverable classes adds a dark background on hover that doesn't exist on prod. Do we want that?

kbangelov (approved):

packages/scratch-gui/src/components/menu-bar/save-status.css
adzhindzhi: It changes the cursor back to the default one (an arrow). Since the `div` element that wraps the `SaveStatus` component uses the `menu-bar-item` class, but we've updated that class to set the pointer to cursor, the text looks like it's clickable when it's not.

Adding the dark background on hover was also intentional, I noticed that it doesn't do that on production, but it makes it more consistent with the other buttons in the navbar. I'm open to suggestions if you have any concerns though.

kbangelov: No yeah, fair enough, seems more consistent that way to me as well.

packages/scratch-gui/src/components/menu-bar/save-status.css
```diff
@@ -2,3 +2,7 @@
     font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
     cursor: pointer;
```
KManolov3: nitpick: this is now redundant for the save now, since it's a button

KManolov3 (approved):

PR #565, opened 2026-05-10
accessibility: Make share & remix buttons more accessible

rosics-code:
### Proposed Changes

Changes remix to `$ui-green-dark` *new color* from `$pen-primary` to match from the project page. Same with share but share in the project page is still yellow.

When a project is shared, the share button will have a `not-allowed` cursor to make sure the person knows it's shared.

### Reason for Changes

Accessibility.

### Test Coverage

Screenshot 2026-05-10 12 46 28 AM
Screenshot 2026-05-10 12 36 32 AM


rosics-code:
@adzhindzhi What do you think?

rubiidev18alt:
I think the previous color was better and actually fit the purple topbar

rosics-code:
we def need this to be accepted

PR #567, opened 2026-05-11
SVG Sandboxing

KManolov3:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-231

### Proposed Changes

- Load SVGs into a sandboxed iframe for measurement vs directly into the DOM.
- Introduce a new function for removing malicious content from SVGs - `canonicalizeSvgText`
- TODO: Route all svg loads through `canonicalizeSvgText` - at the point of `loadVector_` 

### Reason for Changes

Currently we attempt to sanitize SVGs, but the approach is piecemeal. The biggest security issue in the current state is that we load SVGs directly into the DOM, which is an inherently unsafe operation.

### Test Coverage

Added tests for sandboxing, canonicalization and measuring SVGs in a sandboxed environment.


rubiidev18alt:
Please keep the cool css theming via svg 🙏

packages/scratch-render/src/RenderWebGL.js
```diff
@@ -398,16 +398,16 @@ class RenderWebGL extends EventEmitter {
...
-            this._allSkins[skinId].setSVG(svgData, rotationCenter);
-            return;
+            return this._allSkins[skinId].setSVG(svgData, rotationCenter);
         }
 
         const newSkin = new SVGSkin(skinId, this);
-        newSkin.setSVG(svgData, rotationCenter);
         this._reskin(skinId, newSkin);
+        return newSkin.setSVG(svgData, rotationCenter);
     }
```
adzhindzhi, line 411: Now that `createSVGSkin` and `updateSvgSkin` return a promise, if anyone tries to call `getSkinSize` without having awaited the create/update call, the dimensions would not be set yet, right? I think one such instance is the `updateSvg` method in `virtual-machine.js`. We should probably await the update there as well (and anywhere else where it might make sense). Also, looking at the implementation notes, the original plan was to pre-parse the SVG with `DOMParser` to extract `viewBox` synchronously (keeping `_size` available immediately for `getSkinSize` callers), is there a reason that changed?

packages/scratch-svg-renderer/src/sandbox/index.js
```diff
@@ -0,0 +1,181 @@
...
+                this._pendingTickets.set(ticket, {resolve, reject, timeoutId});
+                this._iframe.contentWindow.postMessage(message, '*');
+            });
+        }
+
+        /**
+         * Tear down the iframe and reject any in-flight calls.
+         * After `destroy()`, the next `send()` lazily creates a fresh iframe.
+         */
+        destroy () {
```
adzhindzhi, line 218: Since `destroy()` is currently never called and the `iframe` would stay alive/in memory indefinitely, I wonder if it would make sense for the sandbox to manage its own lifecycle and destroy itself once it becomes idle. For example, maybe the sandbox could start/reset an inactivity timeout whenever a message is sent, and destroy the `iframe` if no new messages arrive within some window (e.g. 10 seconds). That would still keep the sandbox alive during initial project load or bulk SVG imports, while avoiding a permanently alive iframe. I'm mostly thinking out loud here, but it seems like something along these lines could make sense and would avoid coupling the sandbox to the project loading lifecycle?

I'm not sure if there are any concerns performance-wise, but I think it should be relatively cheap, considering that we'd still reuse the iframe during the peak project loading?

packages/scratch-svg-renderer/src/canonicalize-svg.js
```diff
@@ -0,0 +1,228 @@
...
+    name: 'removeDangerousElements',
+    fn: () => ({
+        element: {
+            enter: (node, parentNode) => {
+                if (REMOVE_ELEMENTS.has(node.name)) {
+                    parentNode.children = parentNode.children.filter(
+                        child => child !== node
+                    );
+                    return;
+                }
```
adzhindzhi, line 90: Would the current logic remove tags like `<svg:script>` or `<svg:style>` (or other namespaced tags), which I believe are also valid SVG elements? It seems like they would currently survive the canonicalization step, meaning their CSS rules (including external `url()` references) could persist.

This should be fine for now since `sanitizeSvgText` is still used, and as far as I can tell `DOMPurify` strips those tags (and leaves the content as a text node, which is still not ideal). But if we're planning to deprecate that at some point, or if we want `canonicalize-svg` to handle sanitization more comprehensively, it might be worth addressing here as well.

packages/scratch-svg-renderer/src/canonicalize-svg.js
```diff
@@ -0,0 +1,228 @@
...
+                        delete node.attributes[attr];
+                        continue;
+                    }
+
+                    // Direct URI attributes (href, xlink:href)
+                    if (URI_ATTRS.has(normalizedAttr)) {
+                        const val = node.attributes[attr];
+                        if (val && !isInternalRef(val.replace(/\s/g, ''))) {
+                            delete node.attributes[attr];
+                        }
```
adzhindzhi, line 123: One thing I noticed is that, as far as I can tell, the current canonicalization allows `data:` URIs on all elements, whereas the old sanitization logic only allowed them on image tags. I'm not sure whether this could actually be abused in practice, especially considering the current pipeline, but it still might be worth mentioning.

packages/scratch-svg-renderer/test/snapshots/cat-costume.sanitized.png
adzhindzhi, line 1: Why did this snapshot change, are there any changes in the current sanitization logic that lead to it being changed? It doesn't seem so. Could this be a MacOS vs. Linux difference on how the snapshots are generated, and the cause of the failing tests?

KManolov3: This test was failing for me even before the current changes. It could be an OS difference, now that you mentioned it - will take a look.

packages/scratch-svg-renderer/src/canonicalize-svg.js
```diff
@@ -0,0 +1,228 @@
...
+    } catch {
+        // Unparseable CSS — strip entirely rather than risk external loads.
+        return rawTextHasExternalUrls(ident.decode(cssText)) ? '' : cssText;
+    }
+};
+
+// ── Element / attribute classification ─────────────────────────────────────
+
+/** Elements removed entirely (children discarded). */
+const REMOVE_ELEMENTS = new Set([
```
kbangelov, line 40: I approve of the work that has been done so far on the PR!

The main concern is to not be overly restrictive or not restrictive enough when it comes to removing elements and attributes here. You seem to have taken care of all cases that I could think of. The most important thing here has to be to double-check with a handful of examples right before deploying the code that we catch all of the SVGs that have been known to cause trouble so far and also not restrict the friendly ones (which, by looking at the current code, shouldn't be a thing).

Minor nitpick: consider the `<discard>` element, which might remove some other elements in the SVG (doesn't seem like the end of the world but might consider it while we're at it)

packages/scratch-render/src/RenderWebGL.js
adzhindzhi: I fixed the VM's `updateSvg` to call `updateSVGSkin` asynchronously before reading `getSkinSize`, this was the only affected caller.

On the sync pre-parse: I decided not to do pre-parsing with `DOMParser`. It can only read the declared `viewBox`, but for v2 SVGs that actually need measuring, the `viewBox` is unreliable, so the synchronously returned `_size` would be incorrect, and `normalizeSvg` would overwrite it when it finishes. For SVGs with a valid `viewBox`, measurement is skipped entirely, so there's no real wait to avoid there either (just a microtask), and a pre-parse would parse the SVG twice. So overall I don't think we'd gain much from it.

packages/scratch-svg-renderer/src/sandbox/index.js
adzhindzhi: I implemented an optional inactivity timeout on the sandbox. It now takes an `idleTimeoutMs` option and, if no new messages are sent within that timeout, destroys its iframe. The next send() then recreates it lazily. However, I enabled it only on the measurement sandbox (with a 10s timeout), since that's the one that's permanently alive in almost every session (it's created on project load, even for viewers who never paint). I left the paper sandbox persistent, since it's re-imported on every tab switch (`paper-canvas` unmounts when you leave the Costumes tab), so tearing it down would re-evaluate `Paper.js` on nearly every switch to reclaim only a few MB, which I think could cause a notable delay.

packages/scratch-paint/src/helper/paper-sandbox.js
```diff
@@ -0,0 +1,33 @@
...
+ * Paper.js source is loaded via a dynamic import (code-split chunk) to
+ * avoid doubling the main bundle size.
+ * @returns {Promise<Sandbox>} The Paper.js sandbox instance.
+ */
+const getPaperSandbox = () => {
+    if (!paperSandboxPromise) {
+        paperSandboxPromise = import(
+            /* webpackChunkName: "paper-source" */
+            '@scratch/paper/dist/paper-full.min.js?source'
+        ).then(module => {
```
adzhindzhi, line 19: It's important to note that loading this lazily means we'd need to update scratch-www's webpack configuration to copy the chunk from the package's `dist` folder in `node_modules` into the `js` folder it serves chunks from. It would also mean any other consumer of `scratch-editor` would have to do the same.

packages/scratch-vm/src/import/load-costume.js
```diff
@@ -22,8 +23,9 @@ const loadVector_ = function (costume, runtime, rotationCenter, optVersion) {
         }
 
         // createSVGSkin does the right thing if rotationCenter isn't provided, so it's okay if it's
-        // undefined here
-        costume.skinId = runtime.renderer.createSVGSkin(svgString, rotationCenter);
+        // undefined here.
+        // eslint-disable-next-line require-atomic-updates
+        costume.skinId = await runtime.renderer.createSVGSkin(svgString, rotationCenter);
         costume.size = runtime.renderer.getSkinSize(costume.skinId);
```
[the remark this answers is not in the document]
adzhindzhi: The costume object isn't shared across the async boundary during load, so the rule is a false positive. And assigning it to a local doesn't actually remove the need for `eslint-disable`. Leaving this as-is.

packages/scratch-render/src/RenderWebGL.js
```diff
@@ -355,14 +355,14 @@ class RenderWebGL extends EventEmitter {
...
+     * @returns {Promise<number>} A promise that resolves with the skin ID once the skin's
+     *     dimensions are available via getSkinSize.
      */
     createSVGSkin (svgData, rotationCenter) {
         const skinId = this._nextSkinId++;
         const newSkin = new SVGSkin(skinId, this);
-        newSkin.setSVG(svgData, rotationCenter);
         this._allSkins[skinId] = newSkin;
-        return skinId;
+        return newSkin.setSVG(svgData, rotationCenter).then(() => skinId);
```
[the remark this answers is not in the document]
adzhindzhi: Fixed the in-repo callers. The SVG work is a major release, so we've decided to accept the breaking change rather than preserve backward compatibility. The async return is inherent to the feature - SVG dimensions are now measured through the sandboxed pipeline, so `createSVGSkin`/`updateSVGSkin` genuinely can't hand back a fully-ready skin synchronously.

packages/scratch-paint/src/helper/paper-sandbox.js
```diff
@@ -0,0 +1,33 @@
...
+            const paperSource = module.default;
+            const script = createPaperImportScript(paperSource);
+            return new Sandbox(script);
+        }).catch(err => {
+            // Clear the cached promise so the next call retries rather
+            // than returning the same permanent rejection.
+            paperSandboxPromise = null;
+            throw err;
+        });
```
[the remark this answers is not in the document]
adzhindzhi: That's actually not an issue, the `?source` -> `asset/source` (raw text) mapping comes from the shared `scratch-webpack-configuration` base config, not this package's local `webpack.config.js`. The import returns the `Paper.js` source string as intended, so no local rule or runtime type check is needed.

packages/scratch-svg-renderer/src/canonicalize-svg.js
```diff
@@ -0,0 +1,228 @@
...
+        // get a chance to process it.
+
+        // TODO: Would we want to deny this SVG entirely, instead of letting it passthrough?
+
+        // eslint-disable-next-line no-console
+        console.warn(
+            `canonicalizeSvgText: SVGO could not parse input (${e.reason || e.message}); returning unmodified`
+        );
+        return svgText;
+    }
```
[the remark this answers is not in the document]
adzhindzhi: This is intentional for now. `canonicalizeSvgText` isn't wired into any sanitize/load path yet , so there are no callers relying on it as a security boundary at this point. Deciding the parse-error fallback (passthrough vs. `sanitizeSvgText` vs. deny) is part of the future work to actually integrate canonicalization.

packages/scratch-svg-renderer/src/sandbox/index.js
```diff
@@ -0,0 +1,243 @@
...
+                        // The call settled (by timing out); restart the idle
+                        // countdown if nothing else is in flight.
+                        this._scheduleIdleTeardown();
+                    }, this._timeoutMs);
+                }
+
+                this._pendingTickets.set(ticket, {resolve, reject, timeoutId});
+                this._iframe.contentWindow.postMessage(message, '*');
+            });
+        }
```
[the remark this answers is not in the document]
adzhindzhi: The only trigger for this is a broken injected script, but those are internal, deterministic failures, so re-injecting into the same iframe wouldn't recover them. And resetting the flag on any rejection can't distinguish a script-install failure from other failures, so it would re-inject the script after any errors to handle a case it can't actually fix.

packages/scratch-gui/src/containers/gui.jsx
```diff
@@ -84,6 +89,20 @@ class GUI extends React.Component {
...
+    // Once per editor session, build and warm the Paper.js sandbox during idle
+    // time so the first costume import in the paint editor is not slowed by the
+    // iframe creation and Paper.js evaluation cold-start. Skipped in player-only
+    // sessions, which never open the paint editor.
+    maybePrewarmPaperSandbox () {
+        if (this._paperSandboxWarmed || this.props.isPlayerOnly) return;
+        this._paperSandboxWarmed = true;
+        // Some browsers (e.g. Safari) don't support requestIdleCallback, so fall back to setTimeout.
+        const scheduleIdle = window.requestIdleCallback ?
+            window.requestIdleCallback.bind(window) :
```
KManolov3, line 101: We've likely already checked, but the requestIdleCallback doesn't have anything to do with the Chrome vs Safari performance differences, correct? Paper.js has already loaded by the time we open the Costumes tab in either case?

packages/scratch-svg-renderer/src/load-svg-string.js
```diff
@@ -2,6 +2,36 @@ const SvgElement = require('./svg-element');
...
+ * iframe. Lazily created on the first call to `transformMeasurements`.
+ * @type {Sandbox | null}
+ */
+let measurementSandbox = null;
+
+// Tear the measurement iframe down after this many milliseconds with no
+// measurements. It is created on project load and would otherwise stay in the
+// DOM for the whole session (including player-only sessions that never paint).
+// The window is long enough to span a project load or a burst of costume loads.
+const MEASUREMENT_IDLE_TIMEOUT_MS = 10000;
```
KManolov3, line 20: I am a bit on the fence about destroying sandboxes. I see the value - e.g. it could happen that a previous project load has polluted the sandbox. What worries me is cases where, after the sandbox has been destroyed, initialising a new sandbox ends up adding a few extra seconds on the next project load.

KManolov3: Maybe we could prewarm a new Sandbox, after destroying the old one? Not necessarily insisting on this happening here.

packages/scratch-gui/src/containers/gui.jsx
adzhindzhi: Yeah, it's not related. It generally get's picked up faster on Safari, because it doesn't wait for idle windows, but picks it up immediately after the current task finishes. However, that all happens while the editor is still loading, and especially on large projects, where we fetch multiple assets - all of this is finished long before we navigate to the Costumes tab.

packages/scratch-svg-renderer/src/load-svg-string.js
adzhindzhi: To be fair, I don't think it's worth it. As opposed to the paint sandbox, which takes a lot of time to get created (because of having to pull `paper.js` in as a separate chunk and compile it), the editor sandbox is much lighter to create - tens of milliseconds - so performance there was never really an issue. That's one of the main reasons why I went for destroying only the editor sandbox - recreation is light, it doesn't have a negative impact on performance, and it gives us the benefits of not having a long-lived sandbox. We could technically pre-warm it, tho I'm not sure that would buy us much, we'd basically destroy the iframe and then re-create it right away, so if that's the case, we might as well not destroy it at all. Pre-warming could reduce the creation cost when adding costumes to large v2 projects, but that's all it could save, which itself is not that much - the main slowdown for large projects is the thread being occupied and async functions taking time to start, rather than the actual work.

adzhindzhi: One thing worth mentioning is that currently, library costumes are treated as v2 and always trigger iframe creation and measurements, even for v3 projects - `addCostumeFromLibrary` passes `optVersion = 2`. And since we need to wait for the asset to be fetched before the iframe creation and measurements can start, those tens to a few hundred ms might get noticed. To get around it, though, we'd have to create the sandbox for all projects when the editor mounts and keep it alive for the whole session, which itself might be the worse trade.

PR #574, opened 2026-05-18
ci: automate version and changelog for monorepo releases

cwillisf:
### Proposed Changes

Replace the "draft a GitHub release with a hand-typed version number" flow with a `workflow_dispatch`-triggered pipeline that computes the next version and the changelog from Conventional Commits.

- Hoists `semantic-release` to the monorepo root and adds the `exec` and `git` plugins. Removes the dormant pre-merge `packages/*/release.config.js` files.
- Adds `release.config.js` at the root with this branch model: `develop` (primary), maintenance glob (`N.x` / `N.M.x`), `release/*` with a computed prerelease channel from the suffix, `hotfix/*` likewise. No `@semantic-release/npm`. No `@semantic-release/github`.
- Restructures `publish.yml` into three phases. `release-prep` runs semantic-release with a `pushInsteadOf` redirect to a local bare repository so origin stays untouched. The existing `build` / `test` / `preview` / `cd` jobs consume version and npm tag from `release-prep` outputs. `finalize` pushes the bump commit and tag and creates the GitHub release only after publish succeeds. The `apexskier/github-release-commenter` runs as the last step of `finalize`; it resolves the release via the GitHub API and works fine outside a release-event context.
- Updates `Release.md` to document the new flow.
- Removes `.github/release.yml` (the auto-generate-notes config is unused under the new flow).

### Reason for Changes

- Hand-rolled prerelease identifiers like `13.7.4-svg` proliferate on side branches because nothing enforces semver.
- Changelog quality depends on PR-title hygiene rather than the conventional-commits info already in history.
- The team had this pattern working per-package before the monorepo merge. This work consolidates it for the monorepo shape.

### Test Coverage

Verified locally:

- `npx semantic-release --dry-run --no-ci` on a `release/sr-test` branch (with a local bare repository as `--repository-url`) produces version `13.8.0-sr-test.1` and a changelog from commits since the last tag. All prepare / publish / success plugin steps are correctly skipped.
- Workflow YAML parses cleanly. The `release/*` and `hotfix/*` template-based prerelease channels resolve as designed.

End-to-end verification on a real `release/test-*` branch (with a real npm publish under a throwaway dist-tag, then `npm deprecate`) is intentionally not part of this PR. It needs a real workflow run, which only makes sense once this is on `develop`.


release.config.js
```diff
@@ -0,0 +1,52 @@
+// semantic-release expands `${...}` in its own template syntax, so plain strings
+// here use those references intentionally — the lint rule is off for this file:
+/* eslint-disable no-template-curly-in-string */
+
+// Phase-1 semantic-release config for the scratch-editor monorepo.
```
KManolov3: This may be a leftover from plan notes 😅 I keep meaning to put a global rule to prevent such comments from being added.

Release.md
```diff
@@ -2,16 +2,44 @@
...
+
+If there are no relevant commits since the last release on that branch, the workflow exits cleanly with a
+"nothing to publish" notice. Pressing the button on a branch with only `chore:` or `docs:` commits is a no-op.
+
+## Which branches can release
+
+| Branch                 | Acts as                | npm dist-tag      | Example version          |
+|------------------------|------------------------|-------------------|--------------------------|
+| `develop`              | primary release        | `latest`          | `13.8.0`                 |
+| `13.x`, `13.7.x`, etc. | maintenance            | matching the line | `13.7.5`                 |
```
KManolov3, line 25: Nice!

release.config.js
```diff
@@ -0,0 +1,52 @@
...
+//
+// Runs inside the release-prep job of publish.yml. Produces:
+//   - a version bump applied to every workspace via scripts/npm-version.sh
+//   - a release commit with the changelog embedded in the commit body
+//   - a git tag pointing at the release commit
+//   - .release-version and .release-channel files for the workflow to read
+//     as job outputs (npm_tag is derived from .release-channel: empty = latest)
+//
+// The bump commit and tag are pushed to a local bare repository (configured
+// via --repository-url in the workflow) rather than to origin. The finalize
```
KManolov3: Doesn't it actually end up being configured via `pushInsteadOf`?

KManolov3 (approved):

release.config.js
cwillisf: Good catch! Fixed. Also, good point about the global rule...

release.config.js
cwillisf: Another good catch!

PR #595, opened 2026-05-29
[UEPR-610] make Tips cards navigable through keyboard

adzhindzhi:
### Resolves

[UEPR-610](https://scratchfoundation.atlassian.net/browse/UEPR-610)

### Proposed Changes

- Use actual semantic buttons for the cards header component
- Close card on Esc key pressed

### Reason for Changes

- Tutorial cards cannot be closed through keyboard navigation

[UEPR-610]: https://scratchfoundation.atlassian.net/browse/UEPR-610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/scratch-gui/src/components/cards/card.css
+++ b/packages/scratch-gui/src/components/cards/card.css
@@ -64,6 +64,7 @@ body .card-container {
     height: 44px;
     width: 44px;
     border-radius: 100%;
+    border: none;
     display: flex;
     justify-content: center;
     align-items: center;
@@ -110,6 +111,11 @@ body .card-container {
     border-bottom: 1px solid $extensions-tertiary;
     font-size: 0.625rem;
     font-weight: bold;
+}
+
+.header-buttons button {
+    background: none;
+    border: none;
 }
 
 .header-buttons-hidden {
--- a/packages/scratch-gui/src/components/cards/cards.jsx
+++ b/packages/scratch-gui/src/components/cards/cards.jsx
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import React, {Fragment, useEffect} from 'react';
 import classNames from 'classnames';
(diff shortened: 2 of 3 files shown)


packages/scratch-gui/src/components/cards/card.css
```diff
@@ -64,6 +64,7 @@ body .card-container {
     height: 44px;
     width: 44px;
     border-radius: 100%;
+    border: none;
```
KManolov3, line 67: Does the `.header-buttons button` style already apply the `border: none`, or do we need this as well?

packages/scratch-gui/src/components/cards/cards.jsx
```diff
@@ -370,6 +372,18 @@ const Cards = props => {
     } = props;
     let {x, y} = posProps;
 
+    useEffect(() => {
+        const handleKeyDown = e => {
+            if (e.key === KEY.ESCAPE) {
+                onCloseCards();
```
KManolov3, line 443: Do we want/need to prevent this from bubbling up? I think it should be finе, but asking just in case.

KManolov3 (approved):
Looks good!

packages/scratch-gui/src/components/cards/card.css
adzhindzhi: I'm not sure I understand the question, both `.left-button, .right-button` and `.header-buttons button` apply `border: none` in the current diff. And for the background - `.left-button, .right-button` already have a background set (`background: $extensions-primary;`), that's why we don't need to remove it here. Or am I misunderstanding the question? 😅 

Edit: Oh, sorry, you're asking if we need this at all 😅 Yes, we do, the left/right buttons are not part of the header.

PR #603, opened 2026-06-05
Fix dragging sprites

### LINKED ISSUE #378: Bug when dragging overlapping sprites
https://scratch.mit.edu/discuss/topic/847056/

### Expected Behavior

The topmost sprite is picked up

### Actual Behavior

Depending on the dragging speed, different things can happen:

- the wrong sprite is picked up
- the right sprite is picked up, but the other sprite is hidden and it's motion blocks are disabled

### Steps to Reproduce

Place two sprites such that they are overlapping and the first sprite from the sprite list is on top of the second sprite. Try dragging the first sprite where it covers the second sprite.

If you click and wait until the sprite is automatically picked up, then it usually picks up the second sprite.
If you press the mouse button while the mouse is already moving then it usually hides the second sprite and picks up the first one

1. Go to 'https://scratch.mit.edu/projects/editor/'
2. Add a sprite (eg. 'Bear')
3. Drag and drop the Scratch Cat on the Bear
4. Try dragging the Scratch Cat where it covers the Bear

### System Details
Windows 10, Chrome 132

**Screenshots**

https://github.com/user-attachments/assets/3ce44dfc-570b-44f4-ba30-d1f943f552a6

I've looked into the code and in [stage.jsx#L345-L364](https://github.com/scratchfoundation/scratch-editor/blob/develop/packages/scratch-gui/src/containers/stage.jsx#L345-L364) the draggable targets are tested in the order they appear in `vm.runtime.targets`, which is not the layer order

nimeratus:
### Resolves

- Fixes https://github.com/scratchfoundation/scratch-editor/issues/378

### Proposed Changes

- Remove the data about which sprite is dragged from the `<Stage>`'s React state and put it in its JavaScript properties
- Sort the candidate drawables into layer order in `RenderWebGL.pick()`

### Reason for Changes

- Prevents the situation where it starts dragging a sprite but then the info that it's already dragging something doesn't update before it gets a second mouse event, and so it starts dragging another sprite but doesn't tell the first one that it isn't being dragged anymore
- Makes sure that the topmost sprite is picked up when dragging

### Test Coverage

I've exported [player.html](https://github.com/nimeratus/scratch-editor/blob/9e205cc382dda89dc8b909046e668efa92c3f631/fix-dragging-sprites/scratch-gui/player.html) and [player.js](https://github.com/nimeratus/scratch-editor/blob/9e205cc382dda89dc8b909046e668efa92c3f631/fix-dragging-sprites/scratch-gui/player.js) and imported [this project](https://scratch.mit.edu/projects/1284194741/) into it and tried to reproduce the bugs and they didn't happen

### DIFF
--- a/packages/scratch-gui/src/containers/stage.jsx
+++ b/packages/scratch-gui/src/containers/stage.jsx
@@ -45,12 +45,13 @@ class Stage extends React.Component {
         this.state = {
             mouseDownTimeoutId: null,
             mouseDownPosition: null,
-            isDragging: false,
-            dragOffset: null,
-            dragId: null,
             colorInfo: null,
             question: null
         };
+        this.isDragging = false;
+        this.dragOffset = null;
+        this.dragId = null;
+        
         if (this.props.vm.renderer) {
             this.renderer = this.props.vm.renderer;
             this.canvas = this.renderer.canvas;
@@ -181,7 +182,7 @@ class Stage extends React.Component {
             this.pickY = mousePosition[1];
         }
 
-        if (this.state.mouseDown && !this.state.isDragging) {
+        if (this.state.mouseDown && !this.isDragging) {
             const distanceFromMouseDown = Math.sqrt(
                 Math.pow(mousePosition[0] - this.state.mouseDownPosition[0], 2) +
                 Math.pow(mousePosition[1] - this.state.mouseDownPosition[1], 2)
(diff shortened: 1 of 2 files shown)


nimeratus:
I have read the CLA Document and there was no "not" before the "infringe".
This way that sentence means things like "You warrant that the Contribution will infringe upon the rights of any third party."

rosics-code (approved):
Looks good!

PR #611, opened 2026-06-15
test: report scratch-storage test results in CI

cwillisf:
### Resolves

_No tracking issue; companion to #610 (the same fix for scratch-paint)._

### Proposed Changes

- Add the `jest-junit` reporter to scratch-storage's Jest config, writing to `test-results/` (the path the CI `test-package` action uploads and the `Test Results` job parses).
- Add `jest-junit` to `devDependencies`.
- Remove leftovers from the earlier tap-to-Jest migration: the tap-based `coverage` script and the `tap` config block. Neither worked anymore — `tap` is no longer a dependency — and there are no remaining tap-based tests.
- Ignore `/test-results` in the package `.gitignore`.

### Reason for Changes

scratch-storage runs its tests with Jest, but nothing was configured to emit a JUnit report. The CI `test-package` action only sets `TAP_REPORTER*` env vars (honored by tap, ignored by Jest) and uploads `test-results/*`, so a Jest package with no junit reporter produced no XML and CI had nothing to publish for it. This mirrors the scratch-paint fix in #610.

### Test Coverage

No test changes — this is config only. Verified locally that `npm run test:unit` now writes a well-formed `test-results/junit.xml`, the artifact CI globs via `test-results/**/*.xml`. Confirmed all existing test files already use Jest globals (no tap/tape remnants).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -40437,6 +40437,7 @@
         "eslint-config-scratch": "14.1.30",
         "file-loader": "6.2.0",
         "jest": "29.7.0",
+        "jest-junit": "17.0.0",
         "json": "10.0.0",
         "rimraf": "6.1.3",
         "scratch-webpack-configuration": "3.1.2",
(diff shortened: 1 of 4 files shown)


adzhindzhi (approved):
Looks good to me!

PR #620, opened 2026-06-24
Fix keyboard events dropped after selecting sprite in UI

Chessing234:
## Summary
- Forward keydown events to the VM unless the event target is a text input, textarea, or contentEditable element
- Replaces the previous check that dropped keys for any non-document/SVG target, which caused key-sensing blocks (e.g. "when space key pressed") to stop working after clicking a sprite in the sprite selector

Fixes the same root cause reported in scratchfoundation/scratch-gui#10047 (that repo is archived; this monorepo is the active home for scratch-gui).

## Test plan
- [ ] Add a "when space key pressed" block and confirm space still triggers it after clicking a sprite in the sprite list
- [ ] Confirm Blockly number/text field inputs still accept typing without forwarding keys to the VM
- [ ] Run `npm test -- --testPathPattern=vm-listener-hoc` in `packages/scratch-gui`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/scratch-gui/src/lib/vm-listener-hoc.jsx
+++ b/packages/scratch-gui/src/lib/vm-listener-hoc.jsx
@@ -98,14 +98,9 @@ const vmListenerHOC = function (WrappedComponent) {
             }
         }
         handleKeyDown (e) {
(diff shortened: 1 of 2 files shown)


Chessing234:

PR #662, opened 2026-08-03
Make sure the dist/ folder in scratch-storage is packed

adzhindzhi:
### Resolves

Currently the `/dist` folder in `scratch-storage` is not being packed, because it's globally ignored in the `.gitignore` file on root level, which causes `npm pack` to ignore it by default as well. However, the `/dist` folder includes the `scratch-storage` types, which NGP uses.

When `scratch-storage` was not part of the monorepo, this was fine, because there was no global `.gitignore`, which caused the `/dist` folder to get ignored. 

### Proposed Changes

Add "files" which would get packed both for `scratch-storage` and `scratch-paint`. The same files are packed for the other packages in the monorepo, it looks like this is everything that's needed by consumers.

### Test Coverage

Running` npm pack --dry-run` now includes the `/dist` folder

### DIFF
--- a/packages/scratch-paint/package.json
+++ b/packages/scratch-paint/package.json
@@ -17,6 +17,12 @@
   },
   "main": "./dist/web/scratch-paint.js",
   "browser": "./src/index.js",
+  "files": [
+    "dist",
+    "src",
+    "LICENSE",
+    "TRADEMARK"
+  ],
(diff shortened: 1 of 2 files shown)


cwillisf (approved):
Oops! Thanks 😅

PR #676, opened 2026-08-17
[UEPR-656] fix: always save project thumbnails with a fixed image size

KFilosofov:


packages/scratch-gui/src/lib/store-project-thumbnail.js
```diff
@@ -1,20 +1,46 @@
...
     }
 };
 
-export const getProjectThumbnail = (vm, callback) => {
+export const getProjectThumbnail = (vm, callback, onError) => {
     vm.postIOData('video', {forceTransparentPreview: true});
     vm.renderer.requestSnapshot(dataURI => {
         vm.postIOData('video', {forceTransparentPreview: false});
-        callback(dataURI);
+        scaleToThumbnailSize(dataURI, callback, onError);
```
KManolov3: Can we just do:
````
vm.renderer.requestSnapshot(() => {
        vm.postIOData('video', {forceTransparentPreview: false});
        const canvas = document.createElement('canvas');
        canvas.width = THUMBNAIL_WIDTH;
        canvas.height = THUMBNAIL_HEIGHT;
        canvas.getContext('2d').drawImage(vm.renderer.canvas, 0, 0, THUMBNAIL_WIDTH, THUMBNAIL_HEIGHT);
        callback(canvas.toDataURL());
    });
    vm.renderer.draw();
````
by omitting the Image entirely and using the renderer canvas as source? This way the resize inside requestSnapshot can happen synchronously.

Comment (KManolov3) [reply packages/scratch-gui/src/lib/store-project-thumbnail.js]: I added an example commit here
https://github.com/scratchfoundation/scratch-editor/commit/c717d11de2d220fa0d28d53e8e70b8684312b83a

File: packages/scratch-gui/src/components/stage-header/stage-header.jsx [hunk 2]
```diff
@@ -143,6 +143,11 @@ const StageHeaderComponent = function (props) {
         }
     }, [shouldShowCallout]);
 
+    const handleThumbnailFailure = useCallback(() => {
+        onShowThumbnailError();
+        setIsUpdatingThumbnail(false);
+    }, [onShowThumbnailError]);
+
     const onUpdateThumbnail = useCallback(
         throttle(() => {
```
KManolov3, line 152: I think this throttle doesn't work on NGP currently, because the  `onUpdateProjectThumbnail` callback gets a new reference on every render, because of `storage.saveProjectThumbnail?.bind(storage)` in project-saver-hoc

KManolov3: I don't insist on fixing it now, but maybe we can leave a comment

KManolov3: Opened a PR that should fix most of this - https://github.com/scratchfoundation/scratch-editor/pull/680

packages/scratch-gui/src/lib/store-project-thumbnail.js
KFilosofov: i guess we can :D i wondered the same thing:

> why do we need to convert it to an image? is this the pattern throughout the repo? is this the way people usually resize images?

yet the bot seems to overcomplicate things and defend its decisions good enough. reworked it now

adzhindzhi (approved):
Looks good to me! 🙌

PR #687, opened 2026-08-26
fix(vm): missing check for thread status after executing hats and typo of stackFrames in retireThread

### LINKED ISSUE #685: Stackframe is not cleared in `retireThread`
## Expected Behavior

- `Thread.stackFrames` should be cleared in `Sequencer.retireThread`.
- It should be checked whether the thread is done before invoking `Thread.goToNextBlock`.

## Actual Behavior

In scratch-vm, `Thread.stackFrames` are not cleared correctly in `Sequencer.retireThread` due to typo problems.
https://github.com/scratchfoundation/scratch-editor/blob/a54b373f5e265283ff533d277f00329f9dd8ea03/packages/scratch-vm/src/engine/sequencer.js#L353-L358

It should be `this.stackFrames = [];`.

After fixing this typo problem, it will lead to another problem: project with edge-activated hat blocks will raise error in `reuseStackForNextBlock`. (It can be verified in test/integration/stack-click.js)

```
TypeError: Cannot read properties of undefined (reading 'reuse')

      246 |     reuseStackForNextBlock (blockId: string | null): void {
      247 |         this.stack[this.stack.length - 1] = blockId;
    > 248 |         this.stackFrames[this.stackFrames.length - 1].reuse();
          |                                                       ^
      249 |     }
      250 |
      251 |     /**

      at Thread.reuse [as reuseStackForNextBlock] (src/engine/thread.ts:248:55)
      at Thread.reuseStackForNextBlock [as goToNextBlock] (src/engine/thread.ts:380:14)
      at goToNextBlock (src/engine/runtime.js:2189:24)
          at Array.forEach (<anonymous>)
      at Runtime.forEach [as startHats] (src/engine/runtime.js:2186:20)
      at Runtime.startHats [as _step] (src/engine/runtime.js:2398:22)
      at Timeout._step [as _onTimeout] (src/engine/runtime.js:2973:18)
```

The problem is caused by `Runtime.startHats`. After executing the thread, it's status might be `Thread.STATUS_DONE` and the thread might have been retired, then `thread.stack` and `thread.stackFrame` should be empty array. Then function `reuseStackForNextBlock` will access `this.stack[-1]` and `this.stackFrames[-1]`, and try to call `undefined.reuse()`, which will break the program.

https://github.com/scratchfoundation/scratch-editor/blob/a54b373f5e265283ff533d277f00329f9dd8ea03/packages/scratch-vm/src/engine/runtime.js#L1927-L1930

Previously, since `this.stackFrames` is not cleared and javascript allows assignment to `array[-1]`, vm looks like running well.

## Steps to Reproduce

1. Replace `this.stackFrame = [];` with `this.stackFrames = [];` in function `Sequencer.retireThread`.
2. Run test case `test/integration/stack-click.js`.

## Screenshots

None.

## System Details

None.

alexcui03:
### Resolves

Resolves #685 

- Typo of `thread.stackFrames` in `Sequencer.retireThread`, leading to uncleared stack frames.
- Thread status should be checked before invoking `Thread.goToNextBlock` in `startHats`, since the thread maybe done after calling `execute` and `thead.goToNextBlock()` will raise an error if the above typo is fixed.

### Proposed Changes

- Fix typo of `thread.stackFrames` in `Sequencer.retireThread`.
- Check thread status before calling `thead.goToNextBlock()`. Thread will only go to next block iff it hasn't been done.
- Allow param of `reuseStackForNextBlock` to be null in jsdoc.

### Reason for Changes

Stackframes is not cleared when thread is retired due to typo error. To fix that typo, thread status must be checked before invoking `thread.goToNextBlock()` in `startHats`. Previously, when calling `goToNextBlock`, `thread.stack` is empty and `thread.stackFrames` is not empty, then the following code:

https://github.com/scratchfoundation/scratch-editor/blob/a54b373f5e265283ff533d277f00329f9dd8ea03/packages/scratch-vm/src/engine/thread.js#L256-L259

will execute as normal, since `thread.stack[-1] = null;` is valid in javascript. It will assign null value to property '-1', and `thread.stack.length` will keep 0.

### Test Coverage

pass `vm/integration/stack-click.js`

### DIFF
--- a/packages/scratch-vm/src/engine/runtime.js
+++ b/packages/scratch-vm/src/engine/runtime.js
@@ -1926,6 +1926,10 @@ class Runtime extends EventEmitter {
         // threads are stepped. See ScratchRuntime.as for original implementation
         newThreads.forEach(thread => {
             execute(this.sequencer, thread);
+            if (thread.status === Thread.STATUS_DONE) {
+                // Thread might be done after `execute`.
+                return;
+            }
             thread.goToNextBlock();
         });
         return newThreads;
--- a/packages/scratch-vm/src/engine/sequencer.js
+++ b/packages/scratch-vm/src/engine/sequencer.js
@@ -352,7 +352,7 @@ class Sequencer {
      */
     retireThread (thread) {
         thread.stack = [];
-        thread.stackFrame = [];
+        thread.stackFrames = [];
         thread.requestScriptGlowInFrame = false;
         thread.status = Thread.STATUS_DONE;
     }
--- a/packages/scratch-vm/src/engine/thread.js
+++ b/packages/scratch-vm/src/engine/thread.js
@@ -251,7 +251,7 @@ class Thread {
     /**
      * Reset the stack frame for use by the next block.
      * (avoids popping and re-pushing a new stack frame - keeps the warpmode the same
-     * @param {string} blockId Block ID to push to stack.
+     * @param {string | null} blockId Block ID to push to stack.
      */
     reuseStackForNextBlock (blockId) {
         this.stack[this.stack.length - 1] = blockId;


alexcui03:

PR #691, opened 2026-09-03
fix(vm): resolve dangling variable references the way the runtime does

### LINKED ISSUE #601: bug: variables renamed and local variables appearing on other sprites.
This bug is widespread, data loss causing and, in some cases (see below), project breaking.

---
Imagine waking up to a comment telling you your project is is broken, but you never made any changes to it ... it just spontaneously broke. Well that's what happened to me today. What was weirder was that every remix of the project was broken too. In fact every project on scratch using my code block was broken.

It took me a couple hours but I found the bug: i was iterating over i2, but within the loop using i. This was strange I thought, because it was fine before. Since it was also renamed across all the other projects the only conclusion is Scratch had renamed it not me.

Image

The same bug was apparent across all the broken projects. They had all had i2 renamed to i. In fact in other locations where I had used j2 were renamed.

Image

I believe the bug is to do with local variables appearing in sprites they aren't meant to appear in, triggering a rename. I am unsure at this point. I believe project breaking effects only occur if you use foreach, as the renamed variables still function as independent variables, but foreach works via the variable name not id.

the renaming does not affect turbowarp so it is probably a spork bug.

See one of the remixes of my project (bsp sprite, traversal function) for an example of the renaming: https://scratch.mit.edu/projects/1033253555/
See this one in the render sprite for the local/global variable bug.

cwillisf:
### Resolves

- Resolves scratchfoundation/scratch-editor#601
- Refs scratchfoundation/scratch-editor#533

### Proposed Changes

`Target.reconcileVariableReferences` now resolves a variable, list, or broadcast reference whose id is defined nowhere the same way the runtime's `lookupOrCreateVariable` does when a block executes: by name and type on the sprite first, then on the stage. A sprite-local match wins over a same-named global. Broadcasts match case-insensitively, as the runtime's hat matching does. If nothing in scope has the name, project load creates the variable under that name on the sprite (or on the stage for broadcasts and for references in the stage's own scripts). No name is ever bumped on the project-load path.

Sprite import and backpack paste pass a new `createMissingOnStage` argument to `reconcileVariableReferences`, so a reference still undefined after the by-name lookup is created as a global with a non-colliding name, as before. An exported sprite carries its own locals, so anything left over was a global in the source project.

A new `Runtime.restoreSharedMissingDefinitions` runs before the per-target repair on whole-project load. For any variable or list id referenced from more than one target and defined on no target, it recreates the global on the stage under the referenced name, unless a referencing target owns a local matching its own field name or a same-name global already exists. Fields serialized without an id are skipped. Sprite import does not run it.

The commits are independent and may be easier to review one at a time. Later commits address review feedback.

### Reason for Changes

Since scratchfoundation/scratch-editor#555, whole-project load resolved dangling references against the stage only. A sprite that owned a local `i` and had a `for each` block carrying a stale id but the name `i` got a new global `i2` instead, and only the stale field was rewritten to use it. The loop then counted `i2` while its body read `i`, silently breaking projects (and all their remixes) that had run correctly for years. Before the load-time repair existed, execution-time lookup resolved such fields by name on the sprite first, so this change restores what those projects were built against.

Verified against the project linked from the issue: the stale `for each` fields in the Setup and BSP sprites now resolve to each sprite's own local `i`, and no `i2` global is created.

Not bumping names on the load path keeps name-based access working. A stage script with a dangling `i` while some sprite owns a local `i` used to become `i2`, which broke a sensing block reading "i of Stage". A global and a sprite-local sharing a name is an ordinary Scratch configuration, and scratch-blocks loads both into the workspace without complaint.

The shared-id pre-pass covers the corruption scratchfoundation/scratch-editor#555 was written for. When a global's definition was dropped but several targets still reference its id, they shared one variable, and the shared id is the strongest evidence of that. Repairing each target on its own would give each a private local and silently end that sharing. An id that any target still defines is not a lost global but a stale reference to that target's local, and a same-name local on a referencing sprite is likewise the stronger claim, so the shape reported in scratchfoundation/scratch-editor#601 still resolves locally.

A missing variable whose name carries the cloud prefix is created locally like any other. The prefix is an editor naming convention rather than scope metadata, the loader deliberately keeps cloud-marked sprite-locals as regular locals, and a lost cloud variable used by several sprites is already restored as a global by the pre-pass.

### Test Coverage

New Tap tests in `packages/scratch-vm/test/unit/engine_target.js`:

- a stale reference remaps to a same-name sprite-local, with another sprite also owning an `i` so the old code would have bumped the name
- a same-name local wins over a same-name global
- local creation keeps its name when another sprite has a same-named local, and a second dangling reference to that name coalesces to it
- the same by-name remap through `fixUpVariableReferences`
- a stage script's dangling reference keeps its original name despite a same-named sprite local
- `createMissingOnStage` still bumps past a same-named sprite local
- a dangling broadcast reference matches an existing broadcast case-insensitively
- a cloud-prefixed missing variable stays on the sprite

New Tap tests in `packages/scratch-vm/test/unit/virtual-machine.js`:

- a whole-project load with two sprites plus a global where only the stale field changes
- a missing global shared by several sprites is restored as one global
- a shared missing id is left to per-sprite repair when a sprite owns a same-name local
- a shared missing id remaps to an existing same-name global
- a missing global shared by the stage and a sprite is restored
- sprite import does not run the shared-definition restore
- a shared id that another sprite defines locally is not restored as a global
- a shadowing local is judged by each referring sprite's own field name
- id-less references on different sprites are not merged

Existing tests that asserted stage-side creation on project load were split into project-load and `createMissingOnStage` variants. The `project_*changed_state*` unit tests fail locally on this branch and on `develop` alike because the workspace copy of `@scratch/scratch-storage` has no built `dist`; CI covers them.

https://claude.ai/code/session_01TErdgwd5CzBV3hqX2mseXH

### DIFF
--- a/packages/scratch-vm/src/engine/runtime.js
+++ b/packages/scratch-vm/src/engine/runtime.js
@@ -2595,12 +2595,16 @@ class Runtime extends EventEmitter {
      * the global before each target repairs its own references keeps those targets
      * sharing it rather than each minting a private local.
      *
-     * A same-name local on any referencing target is a stronger claim: that target
-     * would have resolved the reference to its own local by name at execution time,
-     * so the id is left for the per-target repair. Likewise when a same-name global
-     * already exists, since the per-target repair remaps to it. Broadcasts are
-     * skipped because they only ever live on the stage, so every target's repair
-     * already coalesces them there.
+     * An id defined on any target in the project does not qualify, even for
+     * targets that cannot see that definition: it is a stale reference to that
+     * target's local (a script once copied to other sprites), and the runtime
+     * would have given each other sprite its own local. A same-name local on any
+     * referencing target is likewise a stronger claim, judged by that target's own
+     * field name and type: it would have resolved the reference to its own local
+     * by name at execution time, so the id is left for the per-target repair. So
+     * is a same-name global, since the per-target repair remaps to it. Broadcasts
+     * are skipped because they only ever live on the stage, so every target's
+     * repair already coalesces them there.
      *
      * Run before `Target.reconcileVariableReferences` on whole-project load.
      * @param {Array<Target>} targets The targets being installed.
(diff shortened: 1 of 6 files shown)


cwillisf:
Regarding that last Copilot review: the situation it describes has no good answer, so we just need to pick one of the less-bad options from the field of bad answers. The one this PR chooses seems fine. Also, it should be impossible to get into that situation unless you hand-edit the project file, in which case... congrats! You broke it, and you get to keep both pieces.

nimeratus:
> A global and a sprite-local sharing a name is an ordinary Scratch configuration

I'm not sure. If you try to make that using the "Make a Variable" button, it will complain that the variable already exists. In the pre-Spork version of Scratch, the dropdowns and the flyout don't list the local variable if it has the same name as a global one

although it's possible to make it by placing blocks into the Stage that use a global variable, then going to a sprite, deleting and remaking the variable as a local, then running the blocks that use it in the Stage

cwillisf:
> > A global and a sprite-local sharing a name is an ordinary Scratch configuration
> 
> I'm not sure. If you try to make that using the "Make a Variable" button, it will complain that the variable already exists. In the pre-Spork version of Scratch, the dropdowns and the flyout don't list the local variable if it has the same name as a global one
> 
> although it's possible to make it by placing blocks into the Stage that use a global variable, then going to a sprite, deleting and remaking the variable as a local, then running the blocks that use it in the Stage

Fair - maybe "is not always an error" rather than calling it ordinary. :)

